PR #7, opened 2025-04-10
Adding dipole moment functionality

cvmbb:
This addition allows for obtaining the dipole moment from the computed charges.
It also adds a small test for ase, and fixes get_potential_energy() function to return a scalar.


zubatyuk:
Looks good! Thanks for fixing the hard-coded path.

PR #8, opened 2025-05-13
add rocm pipeline

AlbertThie:
Add a separate pipeline to use GPUs on a ROCM environment. This pipeline avoids using gpu-numba, as this has not been updated for ROCM for quite some time. The loading and unloading is a bit hacky, but I haven't noticed too much perfomance degradation. 

Passed all pytest tests.

Tested on a node with 8 AMD Instinct MI250X GPU's.

Please note I did not have access to a NVIDIA  multi GPU environment, some testing may be warranted.

### DIFF
--- a/aimnet/calculators/nb_kernel_rocm.py
+++ b/aimnet/calculators/nb_kernel_rocm.py
@@ -1,0 +1,223 @@
+import numba
+import numpy as np
+
+
+@numba.njit(cache=True, parallel=False, fastmath=True)
+def _nbmat_dual_rocm(
+    coord: np.ndarray,  # float, (N, 3)
+    cutoff1_squared: float,
+    cutoff2_squared: float,
+    mol_idx: np.ndarray,  # int, (N,)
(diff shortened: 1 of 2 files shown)


zubatyuk:
Thanks for the contribution.

Couple of issues:

I notice `.cpu()` and `.cuda()` transfers for the tensors. 
Are these host-device transfers really required?
You should use current device index instead of the first one.

I still need to confirm that the PR does not brake CUDA and CPU compatibility.

isayev:
Closing as stale. The original PR removed gpu-numba, which is no longer used in the codebase. Current main has new NVIDIA-only dependencies (warp-lang for the conv-sv kernel in aimnet/kernels/conv_sv_2d_sp_wp.py, and nvalchemiops for neighbor lists and DFT-D3 in aimnet/calculators/calculator.py and aimnet/modules/{lr,ops}.py) that would need ROCm/HIP backends or fallbacks for AMD GPU support. Reviving the original approach would not address today's blockers. A fresh issue is welcome if there is active demand for ROCm support.

PR #48, opened 2026-03-30
Fix embedded D3TS dispersion detection

k-s-n:
During crystal structure optimization in `AIMNet2Calculator._has_embedded_dispersion()`: for exported models with `has_embedded_lr=True` and `coulomb_mode="sr_embedded"`, it incorrectly returned False, even though the exported model contained an embedded D3TS dispersion module. And so, the calculator could not build the long-range neighbor list (`nbmat_lr` / `shifts_lr`), and `D3TS.forward()` failed with `KeyError: No neighbor matrix found for any suffix in ['_dftd3', '_lr']`.

### DIFF
--- a/aimnet/calculators/calculator.py
+++ b/aimnet/calculators/calculator.py
@@ -441,30 +441,44 @@ class AIMNet2Calculator:
         """
         return self._dftd3_cutoff
 
+#    def _has_embedded_dispersion(self) -> bool:
+#        """Check if model has embedded dispersion (not externalized).
+#
+#        Uses model metadata when available, otherwise returns False (unknown).
+#
+#        Returns
+#        -------
+#        bool
+#            True if model has embedded dispersion module (D3TS or legacy DFTD3).
+#        """
+#        meta = getattr(self.model, "_metadata", None)
+#        if meta is None:
+#            return False  # Unknown, assume no embedded dispersion
+#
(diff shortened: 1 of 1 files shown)


isayev:
Thanks for the catch, @k-s-n! Confirmed the bug: when a v2 `.pt` carries both embedded D3TS and embedded SRCoulomb (`has_embedded_lr=True` + `coulomb_mode='sr_embedded'`), the existing heuristic in `_has_embedded_dispersion()` falls through and returns `False`, breaking crystal optimization with `KeyError`.

I'm closing this PR in favor of #68, which takes a metadata-driven approach instead of runtime `model.modules()` introspection — same pattern just used for `family` / `supports_charged_systems` in #66. Specifically:

- Adds `has_embedded_d3ts: bool` to `ModelMetadata` and propagates it through both load paths (`load_model` + `load_from_hf_repo`).
- `load_v1_model` writes the flag at conversion time using the local variable it already computes.
- `_has_embedded_dispersion()` reads the explicit flag first; the legacy heuristic stays as fallback for pre-flag `.pt` files (no regression).

Two regression tests added — one for the both-set case you reported, one for back-compat. Bug attribution and a link back to this PR are in #68's description.

isayev:
Superseded by #68 (metadata-driven fix; see comment above for rationale).

PR #58, opened 2026-04-24
Long-range Coulomb via nvalchemiops: DSF, Ewald, PME

zubatyuk:
## Summary

- Add nvalchemiops-backed long-range Coulomb backends for DSF, Ewald, and PME. `LRCoulomb` and `AIMNet2Calculator.set_lrcoulomb_method()` now accept `"pme"`, and Ewald/PME use the nvalchemiops default accuracy of `1e-6`.
- Replace the previous long-range custom-op bridge module with direct nvalchemiops calls from `aimnet.modules.lr`. Ewald/PME use nvalchemiops hybrid force/charge-gradient outputs; DSF uses explicit fixed-charge force/virial terms while keeping the energy connected through charges.
- Route external Coulomb and DFT-D3 inference derivatives through a shared `ExternalDerivativeTerms` contract. The calculator adds explicit forces to autograd forces and subtracts the nvalchemiops virial transpose from the row-vector strain gradient before computing stress.
- Support Ewald/PME force and stress training, plus Hessian requests, through local differentiable energy wrappers that keep charge, coordinate, and strain gradients available for double autograd.
- Use nvalchemiops DFT-D3 for detached inference forces/stress and a pure-torch DFT-D3 path for Hessian/double-backward calls. The DFT-D3 strain wrapper remains available for direct differentiable callers.
- Update neighbor-list handling for Ewald/PME to estimate a per-call real-space cutoff from the cell and accuracy, build the required padded flat neighbor list, and keep DSF padded atoms outside cutoff-bounded neighbor rows.
- Update long-range, calculator, model, and tutorial docs for PME, derivative support, DFT-D3 derivative behavior, and the `ewald_accuracy=1e-6` default.

## Behavior Changes

- `LRCoulomb` now supports `simple`, `dsf`, `ewald`, and `pme`; Ewald/PME require periodic `cell` input and estimate their real-space cutoff from `ewald_accuracy` and cell geometry.
- DSF keeps energy differentiable through charges and returns explicit fixed-charge forces/virial for inference, but force/stress training and Hessian requests raise `NotImplementedError`.
- Ewald and PME support inference forces/stress, force/stress losses in training mode, and calculator Hessian requests.
- DFT-D3 returns explicit detached force/virial terms for inference force/stress calls and uses the pure-torch differentiable path for Hessians.
- The external Coulomb and DFT-D3 APIs return `(data, terms)` whenever explicit forces or virial are requested; energy-only calls still return `data`.
- Direct flat PBC Coulomb calls validate the padded input shape contract before dropping the final padding row.
- The `nvalchemi-toolkit-ops` dependency is raised to `>=0.3.1`.

## Testing

- Add calculator tests for Ewald/PME method selection, missing-cell errors, derivative-mode dispatch, combined external Coulomb and DFT-D3 terms, and stress/force/Hessian finite-difference checks.
- Add LR tests for DSF mode-0/mode-2 padding and force behavior, Ewald/PME explicit derivative terms, SR subtraction, charge gradients, and force/stress training paths.
- Add DFT-D3 tests for explicit force/virial terms, embedded differentiable gradients, and molecular/PBC Hessian finite-difference coverage.
- Stabilize the pysis analytic-Hessian smoke test by bounding the RSPRF trust radius and asserting the Hessian path is called.

## Validation

- Local validation: all tests pass locally.


aimnet/modules/lr.py
```diff
@@ -99,13 +127,16 @@ def __init__(
         method: str = "simple",
         dsf_alpha: float = 0.2,
         dsf_rc: float = 15.0,
-        ewald_accuracy: float = 1e-8,
+        ewald_accuracy: float = 1e-5,
```
isayev: Silent BC change from `1e-8`. See top-level review for context; the suggestion below restores the prior default. If the team wants to keep `1e-5`, this needs to be a documented breaking change (CHANGELOG entry + a one-shot `warnings.warn` on first use).

```suggestion
        ewald_accuracy: float = 1e-8,
```

aimnet/calculators/calculator.py
```diff
@@ -600,28 +602,30 @@ def _update_lr_nblists(self) -> None:
 
     def set_lrcoulomb_method(
         self,
-        method: Literal["simple", "dsf", "ewald"],
+        method: Literal["simple", "dsf", "ewald", "pme"],
         cutoff: float = 15.0,
         dsf_alpha: float = 0.2,
-        ewald_accuracy: float = 1e-8,
+        ewald_accuracy: float = 1e-5,
```
isayev: Mirror of the change in `aimnet/modules/lr.py:130` — keep the two defaults in lockstep.

```suggestion
        ewald_accuracy: float = 1e-8,
```

aimnet/modules/lr.py
```diff
@@ -145,34 +176,325 @@ def coul_simple(self, data: dict[str, Tensor]) -> Tensor:
...
+        N = N_padded - 1
+        coord_real = coord[:-1]
+        charges_real = charges[:-1]
+        mol_idx_real = mol_idx[:-1].to(torch.int32)
+        nbmat_real = nbmat[:-1].to(torch.int32)
+        shifts_real = shifts[:-1].to(torch.int32)
+
+        num_systems = int(mol_idx_real.max().item()) + 1
+        # Only pay for virial when the caller is going to take a cell gradient.
+        compute_virial = cell.requires_grad
```
isayev: `compute_virial = cell.requires_grad` couples the virial cost to whatever upstream code happens to do to `cell`.  If a future caller wraps `cell.requires_grad_(True)` for any reason unrelated to stress, the virial is computed (wasted work).  Conversely, a stress request that forgets to set the flag silently gets `compute_virial=False` and `_cell_grad_from_virial` is never invoked — the failure mode is a confusing autograd error downstream rather than a clean assertion here.

Since `_run_external_modules` already knows whether the caller wants stress (line 779 passes `compute_virial=stress` to the DSF path), pass the same flag explicitly through `data` (e.g. `data['_compute_stress'] = stress` set in `eval`) and read it here.  Mirrors the DSF design and removes the implicit channel.

aimnet/calculators/calculator.py
```diff
@@ -726,33 +739,64 @@ def eval(self, data: dict[str, Any], forces=False, stress=False, hessian=False)
...
 
-    def _run_external_modules(self, data: dict[str, Tensor], compute_stress: bool = False) -> dict[str, Tensor]:
-        """Run external Coulomb and DFTD3 modules if attached."""
+    def _run_external_modules(
+        self,
+        data: dict[str, Tensor],
+        *,
+        forces: bool = False,
+        stress: bool = False,
+    ) -> tuple[dict[str, Tensor], LRCoulombDerivativeTerms | None]:
```
isayev: Signature of `_run_external_modules` changed from `(data, compute_stress=False) -> dict` to a tuple-returning method with new kw-only `forces`/`stress` params.  `AIMNet2Calculator` is the public extension point for `aimnet-solvate`, `protonator`, and `LoQI`, all of which subclass it.  Subclasses overriding this method will silently return the old `dict` shape and crash at the unpack on line 754.

Recommend a back-compat shim:

```python
result = self._run_external_modules_impl(data, forces=forces or hessian, stress=stress)
if isinstance(result, dict):
    # Legacy single-return override; assume no explicit Coulomb terms.
    data, coulomb_terms = result, None
else:
    data, coulomb_terms = result
```

and keep the old `compute_stress` kwarg as a deprecated alias for one release.

aimnet/modules/ops.py
```diff
@@ -533,3 +537,387 @@ def dftd3_energy(
...
+    coord_term = torch.zeros_like(target)
+    coord_term.index_add_(0, batch_idx, atom_outer.to(target.dtype))
+
+    residual = target - coord_term
+    if batched_cell:
+        return torch.linalg.solve(cell.mT, residual)
+    return torch.linalg.solve(cell.mT, residual.sum(dim=0))
+
+
+class _LRCoulombFunction(Function):
```
isayev: The dual registration pattern here (`_LRCoulombFunction(Function)` plus the `register_autograd(_lr_coulomb_backward, ...)` at line 843, with the custom op body at line 720 calling `_LRCoulombFunction.apply(...)`) is consistent with the existing DFTD3 pattern in this file (`_DFTD3Function` at line 72, `dftd3_fwd.register_autograd(...)` at line 419), so I'm not flagging it as a defect.

Worth a short module-level docstring explaining the rationale for the dual path (eager goes through `Function.apply`, traced/`torch.compile` goes through `register_autograd`) so future readers don't try to 'simplify' by removing one of the paths and break one of the modes.

aimnet/modules/lr.py
```diff
@@ -115,7 +146,7 @@ def __init__(
         if envelope not in ("exp", "cosine"):
             raise ValueError(f"Unknown envelope {envelope}, must be 'exp' or 'cosine'")
         self.envelope = envelope
-        if method in ("simple", "dsf", "ewald"):
+        if method in ("simple", "dsf", "ewald", "pme"):
             self.method = method
         else:
             raise ValueError(f"Unknown method {method}")
```
isayev, line 294: AIMNet's predicted partial charges almost never sum to integer total charge per molecule.  Ewald/PME are well-defined only for neutral cells; the standard fix is the `tinfoil`/uniform-background correction (which `nvalchemiops` may or may not apply — needs to be verified and documented).

For a charged MOF/zeolite a 0.01 e excess in a small box can shift the energy by hundreds of meV silently.  Two asks:

1. Add a one-time `warnings.warn` in `__init__` (or first `forward`) when `method in ('ewald', 'pme')` and the charge-neutrality assumption is broken at runtime.
2. Add a regression test against an analytic Madelung reference (NaCl rocksalt is canonical).  The current `test_charge_non_neutral_finite` at `tests/test_lr.py:813` only checks finiteness, not numerical correctness.

aimnet/modules/lr.py
```diff
@@ -145,34 +176,325 @@ def coul_simple(self, data: dict[str, Tensor]) -> Tensor:
...
+        # Custom op already converted e²/Å → eV (k_e = Hartree * Bohr) and
+        # summed per-atom → per-system. energies_per_system has shape
+        # (num_systems,) and is float64.
+        e_periodic = energies_per_system
+
         if self.subtract_sr:
-            e = e - self.coul_simple_sr(data)
-        return e
+            data = ops.lazy_calc_dij(data, "")
+            e_periodic = e_periodic - self.coul_simple_sr(data)
```
isayev, line 700: `_calc_coulomb_sr` cuts off at `LRCoulomb.rc=4.6Å` with the bump envelope, but DSF's damping at `α·r = 0.2·4.6 = 0.92` still has `erfc(0.92) ≈ 0.19` — the SR-subtract leaves a residual the NN never trained on.  Likely intentional (you want LR-only to be subtracted), but worth a comment explaining the design.

Also add a fixed-charge dimer regression: for a neutral pair (e.g. Na+/Cl- at 5Å in a 50Å box), `simple` and `dsf` energies should agree to a few meV after the SR subtraction.  No test currently exercises this.

isayev (commented):
Reviewed end-to-end: nvalchemiops backends, the autograd plumbing, the calculator integration, the test suite, and the doc set. The architecture is sound — Ewald/PME via the custom op + register_autograd story is the right shape, the per-call NL build is justified for non-stationary cells, and the test matrix on stress/forces/Hessian is much improved. Below are the items I'd want addressed before merge, then a longer list of land-and-iterate followups.

## Block on fix

1. **Silent BC default change `ewald_accuracy 1e-8 → 1e-5`** (`aimnet/modules/lr.py:130`, `aimnet/calculators/calculator.py:608`).  Existing user code that calls `set_lrcoulomb_method('ewald')` positionally, persisted YAML configs that omit the kwarg, and benchmark scripts pinned to the previous default will all silently shift by ~3 orders of magnitude.  Inline doc snippets in `docs/calculator.md:684,892` and `README.md:131` still quote `1e-8`.  Either restore `1e-8` (suggestion below) or land this as a documented breaking change with a CHANGELOG entry and a single-shot `warnings.warn` the first time the new default is used in inference-only mode.  My preference is restore `1e-8`: per-system Kolafa-Perram error of ~1.4·10⁻⁴ eV per pair contribution at `1e-5` is too loose for force/stress training of dense ionic systems / MOFs (the calculator-FD stress test in `tests/test_pbc.py:1018` uses `5e-3` tolerance precisely because of this).

2. **`_run_external_modules` signature change is subclass-breaking** (`calculator.py:759`).  Old signature `(data, compute_stress=False) -> dict`; new `(data, *, forces=False, stress=False) -> tuple[dict, LRCoulombDerivativeTerms | None]`.  `AIMNet2Calculator` is the public extension point for `aimnet-solvate`, `protonator`, `LoQI`, etc.  Subclasses overriding this method will silently return the old single-`dict` shape and crash at the unpack on line 754.  Recommend adding a thin compatibility shim that detects a 1-tuple/dict return from a subclass override and wraps it as `(data, None)`, deprecate the old shape over one release.

3. **No charge-neutralization handling for non-neutral periodic systems with Ewald/PME** (`lr.py:417`, `tests/test_lr.py:813`).  AIMNet's predicted partial charges almost never sum to integer total charge per molecule.  The existing `test_charge_non_neutral_finite` only checks finiteness, not the physical correctness of the periodic energy.  With the standard `tinfoil`/uniform-background convention, `nvalchemiops` should add a neutralizing background — but this is not asserted, not documented, and not covered by a Madelung-style regression.  For a charged MOF/zeolite a 0.01 e excess in a small box can shift the energy by hundreds of meV.  Add (a) a documentation paragraph in `docs/long_range.md` stating the assumed convention, (b) one regression test against an analytic Madelung reference (NaCl rocksalt is the canonical choice).

## Land-then-followup (open issues; not blocking)

- `compute_virial = cell.requires_grad` (`lr.py:435`) is a brittle implicit-channel; the calculator already knows whether it wants stress.  Plumb an explicit flag through `data` (e.g. `data['_compute_stress']`) or via a per-call kwarg on `LRCoulomb.forward_with_derivatives` mirroring the DSF path.

- `LRCoulombDerivativeTerms` + `forward_with_derivatives` is a parallel public entry point that fires only for DSF; for Ewald/PME it returns `(forward(data), None)`.  Cleaner long-term: every backend returns `LRCoulombDerivativeTerms` (often empty), calculator merges uniformly.  Removes the `external_coulomb.method == 'dsf'` branch in `_run_external_modules`.

- Per-call `AdaptiveNeighborList(cutoff=rs_cut)` instantiation in `calculator.py:946` defeats `AdaptiveNeighborList`'s overflow-buffer caching.  For an MD trajectory where `rs_cut` is essentially constant, cache one per `(_coulomb_method, device)`; invalidate from `set_lrcoulomb_method`.  Same observation for the per-call `estimate_{ewald,pme}_parameters` call.

- DSF and Ewald/PME are never quantitatively cross-checked.  Add `assert_close(E_dsf, E_ewald, rtol=1e-2)` for a small neutral cell with `dsf_rc≥15Å`; add `assert_close(E_pme, E_ewald, atol=accuracy*|E|*10)` for PME parity.  Today only sign-only checks exist.

- DSF stress FD test in `tests/test_pbc.py:188` uses `atol=2e-3, rtol=2e-2`.  DSF is purely real-space and analytically smooth; tighten to ~1e-4 — current tolerance could pass with a stray `.mT` swap on a near-symmetric cell.

- `test_double_backward_smoke` (`tests/test_lr.py:1042`) would pass even if the second-order coord→coord term is silently zero (it is, by design — that's the trade-off for non-differentiable Warp kernels).  The real regression test is `test_force_loss_backward_does_not_raise` at line 962 but it only asserts a non-`None` parameter grad.  Add a numerical FD-of-FD comparison on at least one parameter to lock in the magnitude.

- `_LRCoulombFunction.backward` calls `torch.linalg.solve(cell.mT, residual)` per backward (`ops.py:578-589`).  Numerically dicey for skewed/near-singular cells (slabs, large supercells with one short axis).  For typical MD this runs every gradient step.  Worth a comment justifying the choice; precomputing `cell^{-T}` once via saved tensors is cheap.

- DSF mode-0 padded-atom shift `far = float(self.dsf_rc) * 100.0 + 1.0` (`lr.py:181`) is a magic factor.  For very large structures (solvent slabs at >1500Å along an axis), the shifted padded atoms can land within `dsf_rc` of real atoms.  Use `coord_real.abs().max() + dsf_rc + 1` instead.

- `_coul_nvalchemi` unconditionally drops the trailing row (`lr.py:421`).  Implicit contract that the caller built `_add_padding_row`.  Direct `LRCoulomb(method='ewald')` users who supply their own neighbor list without that pad row will silently lose a real atom.  Add an assert.

- `subtract_sr` semantics with DSF need a comment.  `_calc_coulomb_sr` cuts off at `LRCoulomb.rc=4.6Å`, but DSF's damping at `αr=0.92` (alpha=0.2, r=4.6) leaves `erfc(0.92)≈0.19` undamped — the residual the NN never saw is not zero.  Add a comment + a fixed-charge dimer regression test that `simple` vs `dsf` energies agree at 15 Å for a neutral pair.

- `int(mol_idx_real.max().item())` (`lr.py:432`) forces a GPU→CPU sync on every Coulomb call.  Plumb `num_systems` from the calculator instead.

- Documentation snippet sweep: `docs/calculator.md:684,892`, `README.md:131`, `docs/advanced/charged_systems.md`, and `docs/models/aimnet2_2025.md` still show `ewald_accuracy=1e-8` in examples.  If the new default lands, all of these need updating; if the old default is restored, leave them.

- The dual autograd-registration pattern (`_LRCoulombFunction.apply` inside `lr_coulomb_fwd` *and* `lr_coulomb_fwd.register_autograd(...)` at `ops.py:843`) follows the existing DFTD3 convention in the same file (`_DFTD3Function` at line 72, `dftd3_fwd.register_autograd` at 419), so I'm not flagging it as a defect — eager goes through `Function.apply`, traced/compiled goes through the registered autograd.  Worth a short module-level docstring explaining the dual-mode rationale so future readers don't try to 'simplify' one of the paths.

## CI / merge state

- One conflict: `docs/tutorials/periodic_systems.md` from the admonition-syntax cleanup that landed in #63.  Resolution is mechanical — apply the new admonition convention (no quoted titles) to the new derivative-support paragraph this PR adds.
- `claude-review` job failed on an OIDC `id-token: write` workflow-permission issue, not on the code.
- All other CI green; `tests-gpu` queued.

Nice work overall — the test matrix and the doc reorg are both substantial improvements over the prior Coulomb story.

isayev:
**Resolution for the `docs/tutorials/periodic_systems.md` conflict** (follow-up to the top-level review)

Single hunk, mechanical:

- This branch deletes the inline `!!! warning Ewald summation is currently limited to **single-molecule (single-system) calculations**…` and replaces it with a plain paragraph beginning `DSF supports inference forces and stress, but force/stress losses…`.
- `main` (#63) split that same admonition into the new title + blank-line + 4-space-indented-body form.

**Take this branch's version at the conflict marker** — drop the Ewald-single-system warning, keep the new derivative-support paragraph. The replacement is a regular paragraph, not an admonition, so #63's convention does not apply to it.

The other admonitions in the file (`!!! note When pbc=True…`, `!!! warning The auto-switch warning…`, the two `!!! tip` blocks) are outside this PR's hunks and will three-way-merge cleanly to main's split form — no action needed there.

Verify post-merge with `mkdocs build -s` (should stay clean).

PR #109, opened 2026-07-29
Remove the graph breaks that make torch.compile a net loss

isayev:
`torch.compile` is currently slower than eager on every workload I measured, and the cause is that the model compiles to 42 graphs with 41 breaks — 137 of 144 breaks over five workloads are `Tensor.item()`. A compiled forward re-enters Python about 40 times, and the generated Triton kernels are individually slower than the ATen ones at these sizes, so compilation pays the overhead without ever getting the fusion.

Measured on an L40S, torch 2.9.1+cu128, full `calculator.eval(..., forces=True)`, arms interleaved round-robin in one process, min-of-100 (5 rounds x 20 reps), 4 warmup passes on the first round:

| workload | before | after |
|---|---|---|
| octane B=1 | 0.515x | **1.603x** |
| octane B=8 | 0.847x | **1.771x** |
| octane B=64 | 0.756x | **1.805x** |

On an idle GPU with one arm per process, min-of-25: before 10.99 -> 16.04 ms (0.69x); after 10.33 -> 5.46 ms (1.89x). Graphs 42 -> 4, breaks 41 -> 3. Process-wide over five workloads: frames 225 -> 23, graph breaks 144 -> 12 (all `torch.bincount`), recompile-limit hits 5 -> 0. Kernel launches at B=64 drop 735 -> 326. Cold compile for five workloads: 106 s -> 80 s.

**Eager is bit-for-bit identical.** Every change is gated on `torch.compiler.is_compiling()`. 81 arrays (4 models x 7 workloads x energy/forces/charges) on CPU with `OMP_NUM_THREADS=1`: max |dE|, |dF|, |dq| all exactly 0 against a 1e-9 tolerance. The mode flags stay in the `dict[str, Tensor]` as the eager source of truth, so the TorchScript and serialization contract is unchanged — every edited function still passes `torch.jit.script` (`calc_masks` did not script before this change either, for the reason below).

What changed:

- `nbops.infer_nb_mode` derives the mode from `nbmat`/`numbers` rank — the same definition `set_nb_mode` writes — so the compiled path stops reading `int(data["_nb_mode"].item())`, which ran 27 times per forward.
- `nbops.is_input_padded` returns True under compile. Exact for modes 1 and 2, which always reserve padding; for mode 0 it only means an all-false mask may be applied, and `masked_fill` with an all-false mask is the identity on values and gradients.
- `mol_sum` takes the molecule count from `data["charge"].shape[0]` instead of `int(idx[-1].item())`. That call sat inside a deliberate `is_compiling()` branch whose comment said dynamo handles it — it does, by breaking *and* by syncing device-to-host, since `mol_idx` is on the GPU. Removing this one sync was worth more than every mode flag combined: 11.8 ms -> 5.5 ms at B=64.
- `calc_masks` skips its `data_ptr()`-keyed dedup dict under compile. Dynamo cannot hash a `data_ptr()`, so it was raising an observed exception and **skipping the entire `forward` frame** rather than breaking the graph — 22 compiled frames instead of 9 at B=1. This is also why `calc_masks` was never TorchScript-scriptable.
- `ops.nse` uses the `index_select` gather already sitting next to it instead of `repeat_interleave`, whose output shape is data-dependent.

Two things I expected to matter and measured as not mattering, left alone deliberately: `charges_pre`'s stride alternates (258 from a `.split()` view on pass 0, 1 from a fresh sum after), and `data["charge"]`'s rank alternates 2 then 1 via an internal unsqueeze/squeeze round trip. Both are real, and both only became guarded values because 41 graph breaks made them cross-graph boundaries. Recompiles were 0 before and after. With 4 graphs neither is a boundary, and `charges_pre` is write-only in the whole library — nothing reads it. Adding a `.contiguous()` would have been a fix to a symptom that no longer exists, at a small eager cost.

**A silent inductor miscompile, worth reporting upstream.** Removing the `mol_sum` break let inductor fuse a `scatter_add_` with a downstream gather, and the compiled single-molecule forward returned E = 2.88e21 eV with charges of 4.3e5 — no error, no NaN. Minimal repro:

```python
def f(x, idx, n_out):
    res = torch.zeros(n_out, x.shape[1], device=x.device, dtype=x.dtype)
    res.scatter_add_(0, idx.unsqueeze(-1).expand(-1, x.shape[1]), x)
    res = res + 1.0e-6
    return torch.index_select(res, 0, idx)
```

`torch.compile(f)` against eager on torch 2.9.1+cu128 / L40S gives max |diff| = 2.7 at `n_out=1`, and is correct at 2, 3, 4 and 64. It fires only when the scatter's leading dimension is size 1 and the result is gathered from in the same graph; `index_add_`, out-of-place `index_add`, `.clone()` and `res[idx]` all fail identically, so it is not a mutation-form issue. Worked around here in two compile-gated places: `mol_sum` sums rather than scatters at `out_size == 1`, and `nse` broadcasts rather than gathers for one molecule — `index_select`'s backward is itself a size-1 scatter, and fixing only the forward left forces wrong by 8 eV/A. `test_nse_compiled_matches_eager[1]` fails by 100% on the gradient with that guard removed.

This is the largest risk in the change. If the inductor bug turns out to be broader than "size-1 leading dim gathered in the same graph", these workarounds are patches on a symptom, and the safe fallback is restoring the protective break in `mol_sum` — which costs the entire win.

**The CPU backend keeps the protective break.** CI caught a second member of the same bug family: on the CPU backend through torch 2.10, inductor fuses the `atomic_add` scatters of the PBC distance backward with a dependent pointwise, and `CppScheduling.try_loop_split` fails loudly on the fused group with a degenerate loop split (`AssertionError: expected_var_ranges == extra_indexing_ranges`, `test_torch_compile_pbc_forces`). Both manifestations — the silent CUDA miscompile above and this CPU assertion — are outlawed by the same upstream fusion-legality rule, pytorch/pytorch#172301, first released in 2.11: verified failing identically on CPU with 2.8/2.9.1/2.10 and passing on 2.11/2.12. The fix gates the compiled charge-shape fast path in `mol_sum` to non-CPU tensors, so CPU keeps the `.item()` break — free there, since `.item()` on a CPU tensor syncs nothing — and the graph partitioning that never forms the fused group. Measured wins above are CUDA, where nothing changes on any version. The gate drops when the supported torch floor reaches 2.11.

Compiled-vs-eager fidelity is unchanged by this PR: worst relative force error over 2 models x 5 workloads goes 2.07e-5 -> 1.80e-5, absolute worst 1.6e-4 eV/A on forces of magnitude 16 eV/A. That is fp32 reassociation from different fusion, not a semantic change.

`tests/test_compile_paths.py` adds 12 tests: `infer_nb_mode` equivalent to `set_nb_mode`, `is_input_padded` equivalent to the flag, and compiled-vs-eager parity for `mol_sum` and `ops.nse` including gradients at n_mol in {1, 2, 3, 5}.

Test status: 620 passed, 10 failed. All 10 failures reproduce on clean `main` at 6550e0c and are one upstream cause — `RuntimeError: Trying to backward through alchemiops._batch_ewald_real_space_energy_matrix_backward.default but no autograd formula was registered`, i.e. any second derivative through the Ewald real-space term. Unrelated to this branch.

Two limits worth stating. Only nb_mode 1 was exercised end-to-end, because that is what `AIMNet2Calculator` produces; modes 0 and 2 have unit coverage but no compiled full-model run, and `is_input_padded` returning True under compile is provably the identity for mode 0 rather than empirically confirmed. And `torch.bincount` in `calc_masks` is the last remaining break on CUDA (`mol_sum` keeps one more on CPU, see above) — a statically-sized scatter would remove it, but that creates exactly the size-1 scatter shape inductor miscompiles, so I left it.

### DIFF
--- a/aimnet/models/aimnet2.py
+++ b/aimnet/models/aimnet2.py
@@ -164,7 +164,7 @@ class AIMNet2(AIMNet2Base):
                 _in = torch.cat([self._prepare_in_a(data), self._prepare_in_q(data)], dim=-1)
 
             _out = mlp(_in)
-            if data["_input_padded"].item():
+            if nbops.is_input_padded(data):
                 _out = nbops.mask_i_(_out, data, mask_value=0.0)
 
             if ipass == 0:
--- a/aimnet/modules/core.py
+++ b/aimnet/modules/core.py
@@ -126,7 +126,7 @@ class Output(nn.Module):
 
     def forward(self, data: dict[str, Tensor]) -> dict[str, Tensor]:
         v = self.mlp(data[self.key_in]).squeeze(-1)
-        if data["_input_padded"].item():
+        if nbops.is_input_padded(data):
             v = nbops.mask_i_(v, data, mask_value=0.0)
         data[self.key_out] = v
         return data
--- a/aimnet/nbops.py
+++ b/aimnet/nbops.py
@@ -316,7 +316,7 @@ def mol_sum(x: Tensor, data: dict[str, Tensor]) -> Tensor:
             2,
         ), "Invalid tensor shape for mol_sum, ndim should be 1 or 2"
         idx = data["mol_idx"]
-        if torch.compiler.is_compiling() and "charge" in data:
+        if torch.compiler.is_compiling() and "charge" in data and x.device.type != "cpu":
             # `charge` carries one entry per molecule and is a genuine model
             # input, so its length is static shape metadata: reading it costs
             # no device sync and no graph break.
@@ -325,6 +325,18 @@ def mol_sum(x: Tensor, data: dict[str, Tensor]) -> Tensor:
             # is the same number: mol_sizes comes out of `torch.bincount`, so
             # its length is a data-dependent (unbacked) symbol and sizing an
             # allocation from it hands inductor a shape it cannot reason about.
+            #
+            # CPU is excluded on purpose: it keeps the .item() graph break
+            # below. Through torch 2.10, inductor's CPU scheduler fuses the
+            # atomic_add scatters of the PBC distance backward with a
+            # dependent pointwise, and CppScheduling.try_loop_split then dies
+            # on the fused group with `AssertionError: expected_var_ranges ==
+            # extra_indexing_ranges` (a degenerate loop split). The fusion is
+            # outlawed upstream by pytorch/pytorch#172301, first released in
+            # torch 2.11. The break costs nothing on CPU -- .item() has no
+            # device sync there -- and restores the graph partitioning that
+            # avoids the fused group. Drop this exclusion when the supported
+            # torch floor reaches 2.11.
             out_size = data["charge"].shape[0]
         elif torch.compiler.is_compiling():
             # data dict assembled without `charge`: dynamo handles the .item()
--- a/aimnet/nbops.py
+++ b/aimnet/nbops.py
@@ -16,8 +16,45 @@ def set_nb_mode(data: dict[str, Tensor]) -> dict[str, Tensor]:
     return data
 
 
+def infer_nb_mode(data: dict[str, Tensor]) -> int:
+    """Derive the neighbor mode from tensor metadata alone.
+
+    This is the definition `set_nb_mode` writes into `_nb_mode`, factored out
+    so it can be evaluated without reading a tensor value. Ranks are static
+    metadata, so this costs nothing and -- unlike `Tensor.item()` -- does not
+    force a graph break under `torch.compile`.
+
+    The `numbers` fallback covers data dicts assembled by hand for the packed
+    layout without a neighbor matrix (`mol_sum` only needs `mol_idx`): packed
+    inputs carry a flat 1D `numbers`, dense ones carry (B, N).
+    """
+    if "nbmat" in data:
+        ndim = data["nbmat"].ndim
+        if ndim == 2:
+            return 1
+        if ndim == 3:
+            return 2
+        raise ValueError("Invalid neighbor matrix shape")
+    if "numbers" in data and data["numbers"].ndim == 1:
+        return 1
+    return 0
+
+
 def get_nb_mode(data: dict[str, Tensor]) -> int:
-    """Get the neighbor model."""
+    """Get the neighbor model.
+
+    Eager reads the `_nb_mode` tensor, which is the serialized contract and
+    stays the single source of truth there. Under `torch.compile` the same
+    value is recovered from tensor metadata instead: `Tensor.item()` is a
+    graph break, and this function is called ~27 times per forward, which on
+    its own accounted for most of the breaks in a compiled AIMNet2 forward.
+
+    The two agree by construction for anything that went through
+    `set_nb_mode`, which every model forward calls in `prepare_input` before
+    any consumer runs.
+    """
+    if torch.compiler.is_compiling():
+        return infer_nb_mode(data)
     return int(data["_nb_mode"].item())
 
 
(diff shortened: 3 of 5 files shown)


isayev:
Filed the inductor bug upstream: https://github.com/pytorch/pytorch/issues/191571

The generated Triton kernel confirms the mechanism — the fused kernel loads the gather's value from the scatter target *before* its own `tl.atomic_add` writes it, so the consumer reads the pre-scatter zeros:

```python
tmp8 = tl.load(out_ptr0 + (x0), xmask, eviction_policy='evict_last')   # gather reads here
...
tl.atomic_add(out_ptr0 + (x0), tmp2, xmask, sem='relaxed')             # scatter writes here
tl.store(out_ptr1 + (x2), tmp8, xmask)                                 # stores the stale value
```

Narrowed further than described above: it needs a size-1 leading dimension on the scatter target, a gather from that result in the same graph, **and** CUDA — CPU is exact at both float32 and float64. It is independent of dtype, of the op spelling (`index_add_`, `res[idx]` fail identically), and of intervening pointwise ops. `n_out >= 2` is bit-exact, which suggests the fusion is only attempted when the index is provably a single element.

Reordering alone would not fix it: all program instances in that kernel accumulate into the same few elements with no barrier before the load, so any instance can observe at most a partial sum. That supports keeping the compile-gated workarounds here rather than waiting for a reordering fix.

isayev:
Upstream status update on the inductor miscompile this PR works around.

It is fixed. pytorch/pytorch#172301 landed as `1687bbf` on 2026-02-11, and pytorch/pytorch#191571 was closed as a duplicate of pytorch/pytorch#171493 (same root cause, a PyG-style `index_add_` followed by a gather).

**Version boundary**, checked against the release tags:

| release | date | contains fix |
|---|---|---|
| 2.9.1 (what this branch was measured on) | 2025-11-12 | no |
| 2.10.0 | 2026-01-21 | no |
| **2.11.0** | 2026-03-23 | **yes** |
| 2.12.x, 2.13.0 | | yes |

**The fix is broader than my repro's conditions.** It is not a size-1 special case: `Scheduler.fusable_read_and_write` now returns `False` for any write whose store mode is non-`None` — every `atomic_*` mode plus `tma` — so an atomic scatter is never fused with a dependent read at all. That matches the note in the issue that reordering the load after the `atomic_add` could not have been sufficient (no barrier, and every program instance accumulates into the same elements, so any instance can only observe a partial sum). They prevent the fusion outright rather than reorder it.

**What this means for this PR: nothing changes, keep the workarounds.** `pyproject.toml` allows `torch>=2.8`, so 2.8, 2.9 and 2.10 are all affected and still supported. The two compile-gated formulations in `mol_sum` (sum instead of scatter at `out_size == 1`) and `nse` (broadcast instead of gather at one molecule) are mathematically equivalent, so they cost nothing on 2.11+ and remain load-bearing below it. `test_nse_compiled_matches_eager[1]` stays a meaningful regression test for anyone on an older torch.

**One follow-up is now unblocked, for 2.11+ only.** The last remaining graph break is `torch.bincount` in `calc_masks`. I left it alone because the statically-sized scatter that would replace it constructs exactly the pattern that miscompiled. That constraint is gone on 2.11+, so a version-gated static scatter could take the breaks from 3 to 2. It needs its own measurement — the remaining break may not be worth a version fork in the code — so I am not folding it into this PR.

isayev:
Measured the upstream fix rather than taking it on faith. Isolated venv, torch **2.11.0+cu128** (same CUDA minor as the 2.9.1 this branch was developed on), same idle L40S, this branch checked out unchanged.

**1. The repro, 19 probes, matched pair.** Same script under both interpreters — only the torch version differs.

| | torch 2.9.1 | torch 2.11.0 |
|---|---|---|
| forward, `n_out=1` | 3.43e+00 | **0.0** |
| forward, `n_out` = 2/3/4/8/64 | 0.0 | 0.0 |
| `index_add_` / `res[idx]` spellings at `n_out=1` | 3.43e+00 each | **0.0** |
| CUDA fp64 at `n_out=1` | 3.28e+00 | **0.0** |
| CPU fp32 / fp64 | 0.0 | 0.0 |
| **backward**, scatter→gather, `n_out=1` | **5.48e+01** | **0.0** |
| total | 8 / 19 mismatch | **0 / 19** |

Every failing probe goes to exactly 0.0, not merely under tolerance.

**2. On this branch's actual code, not the toy.** `tests/test_compile_paths.py`, with the two compile-gated workarounds each disabled by hand:

| | torch 2.9.1 | torch 2.11.0 |
|---|---|---|
| workarounds as merged | 12 passed | 12 passed |
| workarounds **disabled** | **1 failed** — `test_nse_compiled_matches_eager[1]`, off by 1.53e+06 | **12 passed** |

So the fix covers the real code path, and 2.9.1 still needs the guard. Worth noting for anyone reading the tests: only the `nse` guard is pinned by a failing test — disabling the `mol_sum` one alone does not fail this suite, since nothing gathers from its result inside the same compiled graph there. It stays as the guard for the full forward, but it is not independently falsified by these tests.

**3. Correction to the graph-break accounting.** I previously described the remaining breaks as bincount. Measured process-wide over the 5 workloads, they are not — and 2.9.1 and 2.11.0 give byte-identical counters:

```
frames total: 23    graph_break total: 12
   6  Dynamic shape operator      aten.bincount.default
   4  Data dependent operator     aten._local_scalar_dense.default
   2  Data-dependent branching
```

**4. The bincount follow-up, tried.** Compile-gated static scatter sized from `charge.shape[0]`, on 2.11:

- graph breaks **12 → 7**, frames **23 → 19** (all six bincount breaks gone)
- compiled-vs-eager fidelity clean: `max |dE| = 0.0 eV`, `max |dF| = 3.3e-05 eV/Å`, `max |dq| = 7.8e-07 e`
- speedup vs eager, two paired runs, interleaved arms, min-of-60, order reversed on the second pair:

| workload | without | with |
|---|---|---|
| octane_B1 | 1.652x, 1.645x | 1.655x, 1.684x |
| octane_B8 | 1.893x, 1.936x | 1.964x, 1.966x |
| octane_B64 | 1.951x, 1.818x | 1.965x, 1.981x |

Consistent direction — every patched value beats every unpatched value in its workload — but the effect is ~1–5%, against a 7% run-to-run spread on B64 alone. **My read: not worth taking now.** It needs a `torch.__version__ >= 2.11` gate to stay correct on the 2.8–2.10 this repo supports, and that is a lot of machinery guarding a correctness-critical branch for a few percent. It becomes a clean two-line change with no gate at all once the floor moves to 2.11, so I would rather wait for that.

Nothing in this PR changes as a result. Its numbers stand and the workarounds stay.

PR #112, opened 2026-08-03
feat(neighbors): support global mode-2 sparse batches

zubatyuk:
# AIMNetCentral Pull Request

## Description

This change adds canonical global-indexed mode-2 sparse neighbor matrices for batches containing molecules of different sizes. It lets callers supply padded batched neighbor lists while preserving correct short-range, long-range, periodic, derivative, and Hessian-vector-product behavior.

Mode-2 inputs are now validated before computation, including padded-center sentinels, per-system ownership, shift alignment, and neighbor-matrix suffix layout. This prevents invalid global indices from reaching CUDA kernels or periodic backends, where failures otherwise can be difficult to diagnose.

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [x] Breaking change (fix or feature that changes existing supported behavior)
- [x] Performance improvement
- [x] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] CI/CD or infrastructure change

## Changes Made

- Added global packed mode-2 neighbor-list handling across calculator, model, AEV, DFT-D3, Coulomb, and periodic long-range paths.
- Validated raw mode-2 neighbor matrices and periodic geometry before flattening, including global index bounds, same-system neighbors, sentinel-only padding, and aligned lattice shifts.
- Added `convert_mode2_local_to_global` for callers migrating legacy local-index 3D matrices to the canonical global representation.
- Preserved shared neighbor-list tensor identity where suffixes alias the same input and avoided re-masking freshly produced mode-1 ConvSV output.
- Made Hessian-vector products select real atoms from the prepared padding mask, so vectors and results are correct for padded mode-2 batches.
- Added CPU, CUDA subprocess, periodic-backend, kernel, calculator, derivative, compile, and documentation coverage for the new layout.

## Compatibility

- Existing flattened mode-1 inputs and canonical global mode-2 inputs remain supported.
- Legacy 3D matrices that use per-system local atom indices must be converted with `aimnet.nbops.convert_mode2_local_to_global` before use.
- Every mode-2 system must reserve its final padded dummy atom; full 3D periodic inputs must supply lattice shifts aligned with each neighbor matrix.

## Testing

- [x] Unit, focused, or full tests pass locally.
- [x] Ruff and repository quality checks pass.
- [x] New tests cover the changed behavior.
- [x] Documentation builds without warnings.

## Checklist

- [x] I have performed a self-review of the code and documentation.
- [x] I have checked that the code and documentation follow the project style.
- [x] I have checked that the documentation is clear and readable.
- [x] I have updated the changelog when the change is user-visible.


isayev:
Reviewed the full diff with empirical verification on an L40S (torch 2.9.1+cu128). The design direction is right — validation-first, clean kernel generalization, good CUDA-subprocess assert testing, and the CUDA compiled mode-2 forward is clean (0 graph breaks, eager parity ~1e-5). Legacy B>1 local-index inputs are rejected loudly on every entry path; no silent-wrong path through the calculator. But one measured physics bug and several majors need fixing before merge.

**Blocker — Ewald/PME energies depend on the padding width.** `_mode2_backend_inputs` builds `batch_idx` over all `B*N` rows (dummies included), and nvalchemiops auto-estimates the Ewald splitting parameter from the per-system atom *count* when only `accuracy` is passed. Padding inflates the count, shifts alpha, and the caller-truncated real-space sum no longer cancels. Measured on a 4-atom cation (12 A cell, `ewald_accuracy=1e-6`, identical geometry and neighbor list): mode-1 flat 1.99822 eV; mode-2 with 1 pad 2.00033; 16 pads 2.00442 — 2.1e-3 eV mode disagreement and 4.1e-3 eV drift across padding widths (~0.1 kcal/mol), unchanged at `accuracy=1e-4`. The same molecule gets a different energy depending on the largest molecule in its batch. Fix: estimate alpha/k-cutoff from real per-system counts (`mol_sizes`) or pass explicit parameters. Note the mandatory trailing dummy makes this bias always nonzero in mode 2.

**Major — shared `(3,3)` cell silently corrupts DSF/DFT-D3 for systems b>0.** `_mode2_backend_inputs` replaced main's `expand(B,-1,-1)` with `unsqueeze(0)`; the batched kernels index `cell[system_id]` unchecked. Measured with two translation-equivalent systems: DSF returns `[-0.8738, -1.3215]` instead of two identical values; DFT-D3 likewise; Ewald/PME at least fail loudly. The calculator path is protected by `normalize_mode2_periodic_geometry`, but `LRCoulomb`/`DFTD3` are public modules driven directly by training code. Restore the expand or validate `cell.shape[0] == B` inside `_mode2_backend_inputs`.

**Major — ConvSV einsum fallback can leak features across systems.** The masked gather index is 0 — system 0's first *real* atom in the flattened layout — so the CPU/float64 branch is only correct because `g_sv` happens to be pre-masked by its producer. Verified: with an unmasked `g_sv`, a real center picks up system-0 features. The CUDA kernel branch is safe (sentinel break). Mask `a_j`/`avf_sv` with `mask_ij` in the fallback rather than relying on a second module's masking discipline; the new `mask_i_` on the output only covers padded centers, not padded neighbor slots of real centers.

**Major — non-finite dummy coordinates poison Ewald** (NaN energy; PME/DSF unaffected). Mode 2 keeps every dummy row in the backend call, unlike mode 1 which strips them. `validate_mode2_nbmat_raw` already asserts shift finiteness — add coordinate finiteness and state the requirement in the mode-2 contract docs.

**Major — `_mode2_check` reintroduces the CPU graph breaks #109 just removed.** `condition.item()` runs inside `models/base.py:prepare_input`, which is inside the compiled forward: a CPU `compile_model=True` mode-2 eval accumulates 20 measured graph breaks (all `Tensor.item()` from nbops), and `fullgraph=True` hard-fails. CUDA is clean — `_assert_async` traces and even survives inductor fusion. Fix: `torch._check(condition)` under `torch.compiler.is_compiling()`, or route the CPU branch through `_assert_async` (it has an eager CPU kernel).

**Major — validation is triplicated and measurably expensive.** The identical 6-line gate is copy-pasted in `calculator.prepare_input`, the `eval` hessian probe, and `models/base.py:prepare_input`, so a plain eval validates twice (three times with `hessian=True`) with no dedup across aliased suffixes: measured 3.6-8.3 ms against a 20.7 ms total eager forward at B=32/N=30/M=60 — 17-40% overhead, launch-bound, with 8 identical validations of the same tensor object. Consolidate to one `nbops.validate_mode2_input()` chokepoint with an identity memo and a validated marker (keeping standalone-model users covered).

**Major — the parity tests cannot see the blocker.** Every periodic parity test compares mode-2 against mode-2 (B=2 vs B=1); the only genuine cross-mode check is non-periodic DSF, which is immune because its parameters are explicit. Please add: mode-1 vs mode-2 Ewald/PME energy/force parity (would have caught the blocker immediately), a padding-width-invariance test (same system, N_real+1 vs N_real+8 padding), a CPU mixed-size periodic case, and a compiled mode-2 forward test asserting zero graph breaks plus eager parity.

**Major — breaking-change bookkeeping.** The semver-breaking input contract gets a single CHANGELOG line under `### Changed` with no BREAKING label. And the break is wider than the "local to global indices" story: every mode-2 input — including previously-unambiguous B=1 — must now reserve `numbers[:, -1] == 0` as a dummy row, which `convert_mode2_local_to_global` explicitly cannot repair (it requires the dummy row to already exist). Label the entry BREAKING and document the dummy-row requirement as an explicit migration step.

Also: the branch now conflicts with main (`CHANGELOG.md`, `aimnet/calculators/calculator.py`, `tests/test_calculator.py`) after #108 merged — needs a rebase regardless.

Suggested order: fix the Ewald/PME parameter estimation first (with the parity + padding-invariance tests pinning it), then the cell expand and the compile-safe check; consolidate validation while touching those lines; rebase; relabel the changelog.

isayev (changes requested):
The global packed mode-2 design is right and the kernel work is high quality — sentinel contract preserved, double-backward capacity fix correct, the no-host-sync regression test is exactly what this path needs. No numerical defects found. Requesting changes on five points that need to land before merge:

1. **Periodic parity tests never leave mode 2.** `_single_mode2_periodic_data` in `tests/test_mode2_periodic_backends.py` builds a dict named `mode1` but unsqueezes into a 3-D nbmat — a mode-2 singleton — and the same pattern repeats in the model/calculator/GPU parity tests, while the one pre-existing cross-mode calculator test was repurposed. A uniform mode-2 bug (shift sign, cell selection) passes the whole suite. Please add one flat mode-1 vs mode-2 anchor per backend (DSF/D3/Ewald/PME) with nonzero shifts — the flat periodic branch in `_coul_nvalchemi` is still there to serve as the reference.

2. **`cutoff_coulomb`/`cutoff_dftd3` are accepted but never consumed or checked.** Mode 2 makes caller-supplied neighbor lists the only path, yet DSF still evaluates with `self.dsf_rc`, D3 with its smoothing radius, Ewald with its accuracy-derived cutoff. A list built with a smaller cutoff passes all structural validation and silently drops pairs — kcal/mol-scale. Consume the keys and check them against the module radii at mode-2 entry (and either validate or hard-document the s ⇒ −s edge-symmetry requirement).

3. **The DSF Hessian claim is untested.** The FD reference in `test_global_mode2_periodic_hessian_diagonal_matches_independent` differentiates the module energy directly, bypassing `calculate_hessian`, and `_periodic_fd_setup` rejects `dsf` — while `docs/long_range.md` now says DSF Hessians work. Since DSF's explicit term is autograd-detached, nothing proves its curvature isn't silently omitted. Add an absolute FD-vs-analytic test through the real calculator path, or change the table to No.

4. **Rebase onto current main needs care beyond the 4 textual conflicts** (CHANGELOG, aev.py, calculator.py, tests/test_calculator.py). Main's AEV dispatch now gates the Warp kernel on `WARP_CUDA_AVAILABLE` (CUDA torch + CPU-only warp-lang is a legal conda-forge solve); the new mode-2 branch checks only device+dtype, so a naive rebase crashes in `wp.launch` on those hosts. Fold the gate + one-time warning into the mode-2 d2features sub-branch, take main's `kernels/__init__.py` and `_init_warp_quietly()` wholesale, and add main's `weights` marker to `tests/test_mode2_cuda_validation.py` (its subprocess loads registry weights). Full suite + GPU run after rebase — `models/base.py` auto-merges cleanly over large unrelated changes, so a clean merge proves nothing.

5. **Hot-path validation cost.** Full mode-2 validation runs in `AIMNet2Calculator.prepare_input`, again in `AIMNet2Base.prepare_input`, and a third time in the Hessian probe — which also does a throwaway `to_input_tensors` device transfer for every Hessian request, mode-2 or not. With the usual all-suffixes-aliased input that's hundreds of extra kernel launches per periodic MD step. Validate once per call chain, memoize by tensor identity, gate the O(B·N·M) content checks behind a trust/debug flag. Fine as a fast-follow if you prefer, but flagging it now since large-system MD is the point of this feature.

isayev:
Picked this up after five weeks without a follow-up: merged `main` and applied the review items on the branch, one commit per item, then ran a second review round over the result and folded its findings in.

**Merge with `main`.** #123 moved PME onto the nvalchemiops energy-graph API and removed the finite-difference Hessian/HVP helpers this branch had extended, so the periodic Coulomb path now keeps `main`'s single energy-only call plus the mode-2 backend-input branch; the dense Hessian keeps the real-atom mask for singleton mode-2 systems and the HVP selects real atoms by index. The two PME parity tests that expected explicit force/virial terms now differentiate the energy like the Ewald ones.

**Ewald/PME padding dependence (the blocker).** The kernel estimates alpha, the k-space cutoff, and the PME mesh from the per-system atom count of the `batch_idx` it receives, and mode 2 hands it every padded row. `_coul_nvalchemi` now runs the kernel's own estimator over the real atoms of the non-empty systems and passes the parameters explicitly. Measured on a 4-atom cation in a 12 A cell (15 A list, accuracy 1e-6, float64): before, mode 2 drifted +2.8e-5 / +4.8e-5 / +5.2e-5 eV from the flat result at 1 / 4 / 16 dummy rows; after, energy, forces, and cell gradient are bit-identical to mode 1 at every padding width for both backends. New tests pin mode 2 to the flat mode-1 evaluation, to itself across padding widths (Ewald/PME through autograd, DSF/DFT-D3 through their explicit terms), and a mixed-size batch to the flat two-molecule batch (the right reference for PME, whose parameters are shared across a batch in both layouts).

**Two contract decisions worth a look.**
- An all-dummy system (a batch padded to a static size) is accepted: it contributes exactly zero energy and is left out of the parameter estimate, so it does not shift the PME alpha of the other systems. Without that, the real-atom estimate gave it `alpha = 0` (NaN for Ewald, a zero division for PME).
- A single `(3, 3)` or `(1, 3, 3)` cell shared by a batch is now broadcast to every system at the normalization layer, the same way a `(3,)` `pbc` already was; the calculator used to reject it while the modules broadcast it. This flips `test_global_mode2_calculator_rejects_shared_cell_for_batch` into a broadcast test (and the `test_calc_distances_single_cell_nb_mode2` check in `test_ops.py` with it); a `(K, 3, 3)` cell with `K` neither 1 nor `B` is still rejected, and `_mode2_backend_inputs` keeps the expand as a defense.

**ConvSV einsum fallback.** Gathered features are zeroed with `mask_ij` instead of relying on a pre-masked `g_sv`; cross-system leak test for both `d2features` settings.

**Validation.** One chokepoint, `nbops.validate_mode2_input`. A plain evaluation validates once: the calculator validates its converted dict and marks it, and the model's `prepare_input` consumes the mark, so a standalone model validates on every call (the mark never survives into a returned dict; the second review caught a first version where it did). A Hessian request on a batch validates the batch once before the split and each re-indexed subsystem once more (3 validations for B=2; the two-suffix probe counted 9 before), with aliased suffixes deduplicated all the way through. Coordinates must be finite, and `_mode2_check` queues `torch._assert_async` when tracing or on CUDA, so the compiled forward has no graph breaks and `fullgraph=True` works; eager CPU keeps `ValueError`, CUDA and compiled code raise `RuntimeError`. Tests cover the validation count per path, alias dedup, mark consumption, finiteness, and a `torch._dynamo.explain` check with zero graph breaks for both the marked and the unmarked entry.

**Coulomb neighbor-list cutoff.** In mode 2 the caller builds `nbmat_coulomb`, so the caller also owns the real-space cutoff the Ewald/PME estimate assumes. The old docs example aliased the model's short-range list into `nbmat_coulomb`, which truncates the real-space sum silently (meV per system at accuracy 1e-6 on a dense cell). The docs now state the requirement with the formula, the example builds a separate list, and the calculator warns when a supplied `cutoff_coulomb` is shorter than its estimate (that key had no consumer before).

**Bookkeeping.** The changelog entry is labeled BREAKING with the dummy-row migration step (the converter cannot add the row), the docs drop the series-internal "PR 03" wording, list the six new optional input keys with their mode-2 shapes, and describe the error semantics (partial PBC is a content check: `ValueError` on eager CPU, `RuntimeError` on CUDA or under compile).

**Left as follow-ups (not blockers):** moving `_split_batch_dim`'s sentinel re-indexing into `nbops`, unifying the four identity memos into one helper, underscoring the nbops validators the tests currently monkeypatch, and a consumer for `cutoff_dftd3`.

Gates run locally on an L40S, since CI has no GPU lane: the GPU-marked mode-2 files, the mode-2 selections across all test modules, the CI-core selection, and the slow mode-2 HVP test are green; `pre-commit run -a` and `mkdocs build -s` are clean. CI needed two follow-up commits. The compile-trace test's skip guard read `torch._dynamo.config.disable`, which mirrors `TORCH_COMPILE_DISABLE`, while the eager-only torch lanes set `TORCHDYNAMO_DISABLE`, which torch checks only at `optimize()` time; the guard now consults both. And the batch-vs-single Hessian-diagonal test took its second-order central difference in float32 with a 1e-3 step, which divides the energy's rounding error by 1e-6: the two translated copies of the same molecule read 0.70 and 0.63 in float32 and 0.7648 for both in float64, and the DSF value was off by half, so the test only passed while both paths rounded bit-identically, which the torch 2.13 CPU wheel did not (0.7315 vs 0.7047). It now runs in float64.

isayev (approved):
Review items from 2026-08-10 are addressed and the follow-up round is folded in. Mode 2 matches the flat evaluation bit for bit, the input validators are fixed, and the Coulomb cutoff guidance now states the real cost. Merging.

isayev:
One more pass before merging, prompted by measuring how large the padding dependence actually was.

**It is tolerance-level.** With a correctly sized Coulomb list, padding moves the energy by at most about 8x the run's own truncation error and forces by at most 9e-6 eV/A, which is one to two float32 ULPs of the energy itself:

| system | energy | own truncation error | worst padding drift | worst force drift |
| --- | --- | --- | --- | --- |
| 4-atom cation, 12 A cell | +2.84 eV | 1.3e-7 eV | 9.7e-7 eV at 31x padding | 2.2e-7 eV/A |
| 192-atom water box | -586.03 eV | 1.7e-5 eV | 1.4e-4 eV | 1.9e-6 eV/A |
| 648-atom water box | -1979.48 eV | 2.2e-4 eV | 1.8e-4 eV | 9.1e-6 eV/A |

The +5.2e-5 eV in my 08-04 note came mostly from the reference side. That repro used a 15 A list where Kolafa-Perram wants 20 A for the cell, and a larger alpha shrinks exactly the real-space tail a short list is missing, so the padded result sat within about 1e-6 eV of converged while the flat reference it was measured against sat 5.2e-5 eV away. The fix is still right, but it earns its place on the contract (mode 1 equals mode 2, kernel parameters tied to physics rather than to padding width) rather than on magnitude. There is now a test pinning that an Ewald system's energy is bit-identical regardless of which companions share its batch.

That reframing moved where the risk actually is, and the re-review found five things worth fixing before merge.

- **The new int32 range check rejected every integer `shifts` tensor.** It compared against `2**31`, which does not fit in int32, so torch wrapped the scalar into the tensor dtype and the comparison was false for every element. int8, int16 and int32 all failed: `ValueError` on CPU, and on CUDA a device-side assert that poisons the context. The dtype allowlist admits those widths and int32 is what the kernels take, so the exposed caller was precisely the mode-2 audience building its own list. Every test used float shifts, which is why this was green.
- **A singular cell gave `nan` instead of an error.** Filling a padded-out slot with a zero cell is the obvious way to write one, and it has no reciprocal lattice; the energy of that system came back `nan` and per-system stress divides by `det(cell)`. Now a content check, with the docs saying to copy a real cell into placeholder slots.
- **The Coulomb cutoff check could not fire where it matters.** It read `cutoff_coulomb` unconditionally while the sum resolves its list through `nbmat_coulomb` or `nbmat_lr`, `cutoff_lr` was not an accepted key at all, and nothing was said when no cutoff was declared, which is the default. It now follows the resolved suffix, accepts `cutoff_lr`, and warns on an undeclared cutoff.
- **The docs understated that trap by three orders of magnitude.** I had written that reusing the model's short-range list costs "on the order of meV per system". True for an isolated small molecule; for a 64-molecule water box with a 5 A list it measures 0.67 eV and 0.07 eV/A against a converged reference. The section now gives both regimes, and notes that the estimated cutoff routinely exceeds half the cell edge for a dilute system, so the list must carry several periodic images and a minimum-image builder cannot satisfy it.
- **An all-dummy batch invented placeholder parameters** where the flat path raises. It now raises too; the constants were unreachable from any test.

Streamlining in the same pass: three clauses of the mode-2 aliasing test compared an object with itself once identity held, PME's alpha reached the kernel as a stride-0 view, and the padding mask was negated twice. The invariants that are load-bearing but invisible (why the neighbor-dtype memo must preserve object identity, why neighbor matrices keep the caller's integer dtype) now say so.

Measured and left as follow-ups: `_real_atom_periodic_parameters` costs 4 host syncs per Ewald call, +0.93 ms against a 3.73 ms kernel, and a bit-identical sync-free formulation exists for Ewald (route dummy rows to a scratch system, slice, substitute with `torch.where`); PME genuinely needs the compaction, because its estimator takes a Python-float median path that raises on a zero count. Validation itself costs about 0.9 ms per evaluation on CUDA with no opt-out, which an MD loop rebuilding the same list shape each step should be able to skip. Also still open: `cutoff_dftd3` has no consumer, and DFT-D3 has the same silent-truncation exposure as the Coulomb list did.

`_mode2_check` was verified on all four paths it has to work on: eager CUDA, `fullgraph=True` on CPU and CUDA, and inside CUDA graph capture and replay. No gradient path changed; the parameters were never differentiable in either mode.

Gates: `pre-commit run -a` and `mkdocs build -s` clean, 672 in the CI-core selection, 671 in the dynamo-disabled lane the torch matrix uses, 48 GPU-marked, 12 slow.

PR #129, opened 2026-09-04
feat(calculators): add harmonic vibrational analysis helper

bkalita23:
Adds `aimnet/calculators/vibrations.py`, a small numpy-only helper for harmonic frequency analysis of the Hessian returned by `AIMNet2Calculator.eval(data, hessian=True)`.

**What it does**

- `analyze_hessian(hessian, positions, masses, *, project_tr=True)`: symmetrizes and mass-weights the Cartesian Hessian, projects out rigid translations and rotations (5 for a linear molecule, 6 otherwise) with a QR-orthonormalized basis, diagonalizes, and returns `VibrationalAnalysis(frequencies_cm1, energies_ev, modes, is_linear, n_tr_removed)`. Imaginary modes are reported as negative wavenumbers; `modes` are unit-norm Cartesian displacements of shape `(n_modes, N, 3)`.
- `masses_amu`, `is_linear_molecule`, `translation_rotation_basis` are exposed as building blocks.
- `vibrational_analysis(calc, data)` wraps the `eval(..., hessian=True)` call and the analysis for a single structure.

**Units and shapes**

- Input Hessian in eV/Å^2 with positions in Å, accepted as `(N, 3, N, 3)` (the `hessian=True` output, see `derivatives.calculate_hessian`) or `(3N, 3N)` (as returned by `AIMNet2ASE.get_hessian`); numpy arrays or torch tensors.
- Output wavenumbers in cm^-1 and quanta in eV, using CODATA 2018 constants. Masses come from the table already bundled in `aimnet.constants`, so the module has no ASE dependency but accepts `atoms.get_positions()` / `atoms.get_masses()` directly and is compatible with the `[ase]` extra.

**Numerical checks (tests/test_vibrations.py, CPU only, no model weights)**

- Diatomic with a known force constant reproduces omega = sqrt(k/mu) to 1e-6 relative (agrees to ~1e-16).
- Water: TR basis Gram matrix equals the identity within 1e-10, 6 rigid-body vectors removed; a mass-weighted isotropic well gives 3 identical real modes; a single negative curvature along a vibrational direction yields exactly one negative frequency.
- CO2 is detected as linear (5 TR vectors, 4 modes); `(N, 3, N, 3)` and `(3N, 3N)` inputs agree.

**Docs**: a short tip in `docs/tutorials/geometry_optimization.md` pointing to the helper from the existing recipe, which drops the six lowest modes and therefore mishandles linear molecules and imaginary frequencies.

### DIFF
--- a/aimnet/calculators/vibrations.py
+++ b/aimnet/calculators/vibrations.py
@@ -1,0 +1,233 @@
+"""Harmonic vibrational analysis of a Cartesian Hessian.
+
+Mass-weights the Hessian, projects out the rigid-body translations and
+rotations (five for a linear molecule, six otherwise), diagonalizes, and
+converts the eigenvalues to wavenumbers. Imaginary modes are reported as
+negative wavenumbers, so they are never confused with the near-zero
+rigid-body block that the ``sorted(freqs)[6:]`` recipe relies on.
+
+The input is the dense Hessian returned by
+``AIMNet2Calculator.eval(data, hessian=True)["hessian"]`` -- shape
+``(N, 3, N, 3)`` in eV/Å^2 for coordinates in Å, assembled by
+:func:`aimnet.calculators.derivatives.calculate_hessian` -- or its
+``(3N, 3N)`` reshape as returned by
+:meth:`aimnet.calculators.aimnet2ase.AIMNet2ASE.get_hessian`.
+
+Only numpy and the mass table bundled in :mod:`aimnet.constants` are used, so
+the module works with or without the ``ase`` extra; ASE users can pass
+``atoms.get_positions()`` and ``atoms.get_masses()`` directly.
+"""
+
+import math
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, Any
+
+import numpy as np
+import torch
+
+from aimnet.constants import get_masses
+
+if TYPE_CHECKING:
+    from .calculator import AIMNet2Calculator
+
+# CODATA 2018 recommended values (https://physics.nist.gov/cuu/Constants/).
+# e, h and c are exact since the 2019 SI revision.
+_E = 1.602176634e-19  # elementary charge, C
+_HBAR = 6.62607015e-34 / (2.0 * math.pi)  # reduced Planck constant, J s
+_AMU = 1.66053906660e-27  # unified atomic mass unit, kg
+_C = 299792458.0  # speed of light in vacuum, m/s
+
+# hbar*omega in eV for a mass-weighted Hessian eigenvalue omega^2 in eV A^-2 amu^-1.
+_EV_PER_SQRT_EIGENVALUE = _HBAR * 1e10 / math.sqrt(_E * _AMU)
+# Energy of a photon with wavenumber 1 cm^-1, in eV (``ase.units.invcm``).
+_INVCM = 100.0 * 2.0 * math.pi * _HBAR * _C / _E
+
+
+@dataclass
+class VibrationalAnalysis:
+    """Result of :func:`analyze_hessian`.
+
+    Attributes:
+        frequencies_cm1: Harmonic wavenumbers in cm^-1, shape ``(n_modes,)``, in ascending order of the
+            mass-weighted Hessian eigenvalue. Imaginary modes are reported as negative values.
+        energies_ev: Vibrational quanta ``h*nu`` in eV with the same ordering and sign convention.
+        modes: Unit-norm Cartesian displacement vectors, shape ``(n_modes, N, 3)``.
+        is_linear: Whether the geometry was detected as linear.
+        n_tr_removed: Number of rigid-body modes projected out (0, 5 or 6).
+    """
+
+    frequencies_cm1: np.ndarray
+    energies_ev: np.ndarray
+    modes: np.ndarray
+    is_linear: bool
+    n_tr_removed: int
+
+    @property
+    def n_imaginary(self) -> int:
+        """Number of imaginary modes (negative wavenumbers)."""
+        return int(np.sum(self.frequencies_cm1 < 0))
+
+
+def masses_amu(atomic_numbers: Any) -> np.ndarray:
+    """Standard atomic masses in amu for ``atomic_numbers``.
+
+    Uses the table bundled in :func:`aimnet.constants.get_masses` (the values of
+    ``ase.data.atomic_masses`` stored in single precision), so no ASE install is
+    required. Pass ``atoms.get_masses()`` to :func:`analyze_hessian` instead when
+    ASE is available and full double precision matters.
+
+    Args:
+        atomic_numbers: Integer array-like of shape ``(N,)``.
+
+    Returns:
+        Array of shape ``(N,)`` with dtype float64.
+    """
+    numbers = np.asarray(atomic_numbers, dtype=np.int64).reshape(-1)
+    return get_masses().double().numpy()[numbers]
+
+
+def is_linear_molecule(positions: np.ndarray, masses: np.ndarray, tol: float = 1e-4) -> bool:
+    """Return ``True`` when the smallest principal moment of inertia is negligible.
+
+    Args:
+        positions: Cartesian coordinates, shape ``(N, 3)``, in Å.
+        masses: Atomic masses, shape ``(N,)``, in amu.
+        tol: Ratio of the smallest to the largest principal moment below which
+            the molecule is treated as linear. Atoms and diatomics are always linear.
+    """
+    positions = np.asarray(positions, dtype=np.float64).reshape(-1, 3)
+    masses = np.asarray(masses, dtype=np.float64).reshape(-1)
+    if len(masses) <= 2:
+        return True
+    r = positions - (positions * masses[:, None]).sum(axis=0) / masses.sum()
+    inertia = np.eye(3) * np.sum(masses * np.einsum("ij,ij->i", r, r)) - np.einsum("i,ij,ik->jk", masses, r, r)
+    moments = np.sort(np.linalg.eigvalsh(inertia))
+    return bool(moments[0] < tol * max(moments[-1], 1e-12))
+
+
+def translation_rotation_basis(positions: np.ndarray, masses: np.ndarray, is_linear: bool) -> np.ndarray:
+    """Orthonormal basis of rigid-body motions in mass-weighted Cartesian coordinates.
+
+    The three translations and three infinitesimal rotations about the center of
+    mass are scaled by ``sqrt(m)`` and orthonormalized with a QR decomposition.
+    For a linear molecule the rotation about the molecular axis is a null
+    vector and is dropped.
+
+    Args:
+        positions: Cartesian coordinates, shape ``(N, 3)``, in Å.
+        masses: Atomic masses, shape ``(N,)``, in amu.
+        is_linear: Result of :func:`is_linear_molecule`.
+
+    Returns:
+        Array of shape ``(k, 3N)`` with orthonormal rows, ``k = 5`` for a linear
+        molecule and ``6`` otherwise (fewer only for a single atom).
+    """
+    positions = np.asarray(positions, dtype=np.float64).reshape(-1, 3)
+    masses = np.asarray(masses, dtype=np.float64).reshape(-1)
+    n = len(masses)
+    sqm = np.sqrt(masses)
+    r = positions - (positions * masses[:, None]).sum(axis=0) / masses.sum()
+    vectors = []
+    for k in range(3):
+        t = np.zeros((n, 3))
+        t[:, k] = sqm
+        vectors.append(t.ravel())
+    for k in range(3):
+        axis = np.zeros(3)
+        axis[k] = 1.0
+        vectors.append((np.cross(axis, r) * sqm[:, None]).ravel())
+    q, rr = np.linalg.qr(np.array(vectors).T)  # columns are the orthonormalized vectors
+    diag = np.abs(np.diag(rr))
+    q = q[:, diag > 1e-6 * max(diag.max(), 1e-12)]
+    expected = 5 if is_linear else 6
+    return q.T[:expected]
+
+
+def _as_square_hessian(hessian: Any, n: int) -> np.ndarray:
+    """Return ``hessian`` as a float64 ``(3N, 3N)`` array, accepting the ``(N, 3, N, 3)`` layout too."""
+    if isinstance(hessian, torch.Tensor):
+        hessian = hessian.detach().cpu().numpy()
+    h = np.asarray(hessian, dtype=np.float64)
+    if h.shape not in ((3 * n, 3 * n), (n, 3, n, 3)):
+        raise ValueError(f"Expected a Hessian of shape (3N, 3N) or (N, 3, N, 3) with N={n}, got {h.shape}")
+    return h.reshape(3 * n, 3 * n)
+
+
+def analyze_hessian(
+    hessian: Any,
+    positions: np.ndarray,
+    masses: np.ndarray,
+    *,
+    project_tr: bool = True,
+) -> VibrationalAnalysis:
+    """Harmonic frequencies and normal modes from a Cartesian Hessian.
+
+    The Hessian is symmetrized, mass-weighted, optionally projected onto the
+    complement of the rigid-body translations and rotations, and diagonalized.
+    With projection the ``n_tr_removed`` eigenvalues closest to zero are the
+    rigid-body null space and are discarded; the remaining modes are returned
+    in ascending order of eigenvalue, so imaginary modes (negative eigenvalue,
+    reported as negative wavenumber) come first.
+
+    Args:
+        hessian: Cartesian Hessian in eV/Å^2 with shape ``(3N, 3N)`` or ``(N, 3, N, 3)``, as a numpy
+            array or torch tensor. This is the layout and unit of
+            ``AIMNet2Calculator.eval(data, hessian=True)["hessian"]``.
+        positions: Cartesian coordinates, shape ``(N, 3)``, in Å.
+        masses: Atomic masses, shape ``(N,)``, in amu (see :func:`masses_amu`).
+        project_tr: Project out translations and rotations. When ``False`` all ``3N`` modes are
+            returned, including the near-zero rigid-body ones.
+
+    Returns:
+        :class:`VibrationalAnalysis` with ``3N - n_tr_removed`` modes.
+    """
+    positions = np.asarray(positions, dtype=np.float64).reshape(-1, 3)
+    masses = np.asarray(masses, dtype=np.float64).reshape(-1)
+    n = len(masses)
+    if positions.shape[0] != n:
+        raise ValueError(f"positions has {positions.shape[0]} atoms but masses has {n}")
+    h = _as_square_hessian(hessian, n)
+    h = 0.5 * (h + h.T)
+    inv_sqrt_m = np.repeat(1.0 / np.sqrt(masses), 3)
+    h_mw = h * inv_sqrt_m[:, None] * inv_sqrt_m[None, :]
+    linear = is_linear_molecule(positions, masses)
+    removed = 0
+    if project_tr and n > 1:
+        basis = translation_rotation_basis(positions, masses, linear)
+        projector = np.eye(3 * n) - basis.T @ basis
+        h_mw = projector @ h_mw @ projector
+        h_mw = 0.5 * (h_mw + h_mw.T)
+        removed = basis.shape[0]
+    omega2, vecs = np.linalg.eigh(h_mw)
+    if removed:
+        order = np.argsort(np.abs(omega2))
+        keep = np.sort(order[removed:])  # drop the null space, keep ascending eigenvalue order
+        omega2, vecs = omega2[keep], vecs[:, keep]
+    energies = _EV_PER_SQRT_EIGENVALUE * np.sqrt(np.abs(omega2)) * np.where(omega2 < 0, -1.0, 1.0)
+    modes = (vecs * inv_sqrt_m[:, None]).T.reshape(-1, n, 3)
+    norms = np.linalg.norm(modes.reshape(len(modes), -1), axis=1)
+    modes = modes / np.where(norms > 0, norms, 1.0)[:, None, None]
+    return VibrationalAnalysis(
+        frequencies_cm1=energies / _INVCM,
+        energies_ev=energies,
+        modes=modes,
+        is_linear=linear,
+        n_tr_removed=removed,
+    )
+
+
+def vibrational_analysis(
+    calc: "AIMNet2Calculator", data: dict[str, Any], *, project_tr: bool = True
+) -> VibrationalAnalysis:
+    """Compute the Hessian of a single structure with ``calc.eval(data, hessian=True)`` and analyze it.
+
+    Args:
+        calc: An :class:`~aimnet.calculators.AIMNet2Calculator`.
+        data: Single-structure input as accepted by :meth:`AIMNet2Calculator.eval`, with ``coord`` of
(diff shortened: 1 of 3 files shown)


aimnet/calculators/vibrations.py
```diff
@@ -0,0 +1,233 @@
...
+        vectors.append(t.ravel())
+    for k in range(3):
+        axis = np.zeros(3)
+        axis[k] = 1.0
+        vectors.append((np.cross(axis, r) * sqm[:, None]).ravel())
+    q, rr = np.linalg.qr(np.array(vectors).T)  # columns are the orthonormalized vectors
+    diag = np.abs(np.diag(rr))
+    q = q[:, diag > 1e-6 * max(diag.max(), 1e-12)]
+    expected = 5 if is_linear else 6
+    return q.T[:expected]
```
isayev, line 143: **Blocking.** The rigid-body basis is wrong for any axis-aligned linear molecule that is not *exactly* linear — which is every optimized or file-parsed geometry — and the result depends on the lab orientation.

Two regimes, one root cause (unpivoted QR + positional slice):

1. Below the `1e-6` mask: when the molecular axis is along *x* or *y*, the near-null axial-rotation column comes third. Householder QR *normalizes* it into a noise direction and orthogonalizes `r_y`/`r_z` against it before the mask discards it. The surviving five vectors are orthonormal but do not span the rotation space.
2. Between `1e-6` here and the `1e-4` inertia tolerance in `is_linear_molecule`: all six columns survive and `[:5]` drops the *last* one — a genuine rotation.

Measured with the `aimnet2` registry model on CO2, 0.05° off-axis, same geometry rotated:

```
axis z:  [ 680.2  680.2  1373.7  2401.3 ]   correct, degenerate bend
axis x:  [  53.0  680.2  1373.7  2401.3 ]   one bend lost, spurious 53 cm^-1 mode
```

With 1e-6 Å noise the wrong bend lands at 547.67 or 61.41 cm^-1 (true 675.25); after a BFGS optimization to fmax 1e-4 it came out at 224.4 in one run and 301.4 in the next. In those cases the helper is worse than the `sorted(freqs)[6:]` recipe it replaces, which gets CO2 right. Exact-zero geometries pass only because LAPACK sets tau=0 for an exactly-zero column.

SVD fixes both regimes in one step: singular values come out descending, so keeping the leading columns always discards the smallest direction (the axial rotation) whatever the orientation, and `is_linear` becomes the single rank authority — no second threshold to disagree with it. The `n == 1` case gives the three translations (needed for the single-atom fix below).

```suggestion
    # SVD, not QR: singular values come out descending, so keeping the leading
    # columns always discards the near-null axial rotation of a (near-)linear
    # molecule, whatever its orientation. Unpivoted QR normalizes that tiny
    # column into noise and contaminates the rotations that follow it.
    u, _, _ = np.linalg.svd(np.array(vectors).T, full_matrices=False)
    expected = 3 if n == 1 else (5 if is_linear else 6)
    return u[:, :expected].T
```

Verified: x, y and z all give `[680.2 680.2 1373.7 2401.3]`; caffeine unchanged; all nine existing tests pass. Please add the regression test described on the test file — the current suite cannot see this.

aimnet/calculators/vibrations.py
```diff
@@ -0,0 +1,233 @@
...
+        atomic_numbers: Integer array-like of shape ``(N,)``.
+
+    Returns:
+        Array of shape ``(N,)`` with dtype float64.
+    """
+    numbers = np.asarray(atomic_numbers, dtype=np.int64).reshape(-1)
+    return get_masses().double().numpy()[numbers]
+
+
+def is_linear_molecule(positions: np.ndarray, masses: np.ndarray, tol: float = 1e-4) -> bool:
```
isayev, line 89: With the SVD change this becomes the *only* place that decides linear vs. not, which is the right shape. Worth a sentence in the docstring saying so, and that `tol=1e-4` corresponds to roughly 1° off-axis. A molecule flagged non-linear that is in fact linear would have a null sixth vector projected out of the vibrational space, so the generous default is the safe direction.

aimnet/calculators/vibrations.py
```diff
@@ -0,0 +1,233 @@
...
+    n = len(masses)
+    if positions.shape[0] != n:
+        raise ValueError(f"positions has {positions.shape[0]} atoms but masses has {n}")
+    h = _as_square_hessian(hessian, n)
+    h = 0.5 * (h + h.T)
+    inv_sqrt_m = np.repeat(1.0 / np.sqrt(masses), 3)
+    h_mw = h * inv_sqrt_m[:, None] * inv_sqrt_m[None, :]
+    linear = is_linear_molecule(positions, masses)
+    removed = 0
+    if project_tr and n > 1:
```
isayev, line 195: Minor, and only together with the `reshape` change below: for a single atom this guard skips projection, so an N=1 Hessian with float32 noise returns three ~0 cm^-1 "modes" with `n_tr_removed=0`, contradicting the docstring (`0, 5 or 6`). With the SVD basis returning 3 rows for N=1, the guard can go:

```suggestion
    if project_tr:
```

Applied alone this crashes on N=1 at the `reshape(len(modes), -1)` below (ambiguous `-1` on an empty array) — apply both, and update the docstring to `0 when not projected; 3, 5 or 6 otherwise`.

aimnet/calculators/vibrations.py
```diff
@@ -0,0 +1,233 @@
...
+        h_mw = 0.5 * (h_mw + h_mw.T)
+        removed = basis.shape[0]
+    omega2, vecs = np.linalg.eigh(h_mw)
+    if removed:
+        order = np.argsort(np.abs(omega2))
+        keep = np.sort(order[removed:])  # drop the null space, keep ascending eigenvalue order
+        omega2, vecs = omega2[keep], vecs[:, keep]
+    energies = _EV_PER_SQRT_EIGENVALUE * np.sqrt(np.abs(omega2)) * np.where(omega2 < 0, -1.0, 1.0)
+    modes = (vecs * inv_sqrt_m[:, None]).T.reshape(-1, n, 3)
+    norms = np.linalg.norm(modes.reshape(len(modes), -1), axis=1)
```
isayev, line 208: Companion to the single-atom change (harmless on its own): an explicit second dimension lets an empty `modes` array reshape.

```suggestion
    norms = np.linalg.norm(modes.reshape(len(modes), 3 * n), axis=1)
```

Verified: single atom → `n_tr_removed=3`, 0 modes, `modes.shape == (0, 1, 3)`; H2 diatomic → 5 removed, one mode at 2322.8 cm^-1 = analytic.

aimnet/calculators/vibrations.py
```diff
@@ -0,0 +1,233 @@
...
+    Args:
+        calc: An :class:`~aimnet.calculators.AIMNet2Calculator`.
+        data: Single-structure input as accepted by :meth:`AIMNet2Calculator.eval`, with ``coord`` of
+            shape ``(N, 3)`` in Å and ``numbers`` of shape ``(N,)``.
+        project_tr: Passed to :func:`analyze_hessian`.
+    """
+    coord = torch.as_tensor(data["coord"]).detach().cpu().numpy().reshape(-1, 3)
+    numbers = torch.as_tensor(data["numbers"]).detach().cpu().numpy().reshape(-1)
+    hessian = calc.eval(data, hessian=True)["hessian"]
+    return analyze_hessian(hessian, coord, masses_amu(numbers), project_tr=project_tr)
```
isayev, line 233: Two things in the wrapper:

**Embedded-dispersion artifacts (should at least warn).** For a model that *embeds* a `DFTD3` module, the calculator's Hessian currently omits the dispersion curvature entirely — `_DFTD3EnergyFunction.backward` returns graph-free `-forces * g`, and `calculate_hessian` passes `allow_unused=True`, so nothing raises. No registry model embeds D3 (they externalize it and `hessian=True` takes the twice-differentiable torch path), but some artifacts do, and this helper is exactly where users will consume the result. Proxy magnitude on the external-D3 model: ~6 cm^-1 on caffeine's lowest mode, ~8 cm^-1 on a benzene dimer's intermolecular mode — worst where people want frequencies of weakly bound systems. Fixing the calculator is a separate item on my side; until then this function should say so:

```python
if getattr(calc, "_has_embedded_dispersion", lambda: False)():
    warnings.warn(
        "This model embeds a dispersion module whose curvature the calculator's Hessian "
        "currently omits; low-frequency modes of dispersion-bound systems will be off.",
        stacklevel=2,
    )
```

(`getattr` keeps the `FakeCalculator` in the tests working; needs `import warnings`.)

**Batched input.** A `(B, N, 3)` coordinate with `B > 1` gets flattened to `(B*N, 3)` here and fails downstream with `Expected a Hessian ... with N=48` — loud but misleading. Guard up front:

```python
coord = torch.as_tensor(data["coord"]).detach().cpu()
if coord.ndim == 3 and coord.shape[0] > 1:
    raise ValueError("vibrational_analysis handles one structure at a time; loop over the batch")
```

`(N, 3)` and `(1, N, 3)` both work as-is. A `masses=None` keyword override would also be welcome for isotope work; `atoms.get_masses()` users currently have to call `analyze_hessian` themselves.

aimnet/calculators/vibrations.py
```diff
@@ -0,0 +1,233 @@
...
+    ASE is available and full double precision matters.
+
+    Args:
+        atomic_numbers: Integer array-like of shape ``(N,)``.
+
+    Returns:
+        Array of shape ``(N,)`` with dtype float64.
+    """
+    numbers = np.asarray(atomic_numbers, dtype=np.int64).reshape(-1)
+    return get_masses().double().numpy()[numbers]
```
isayev, line 86: Z=0 (the calculator's padding value) has mass 0 here and surfaces later as `LinAlgError: Eigenvalues did not converge`; Z=-1 silently wraps to Og. Validate the range:

```suggestion
    numbers = np.asarray(atomic_numbers, dtype=np.int64).reshape(-1)
    table = get_masses().double().numpy()
    if numbers.size and (numbers.min() < 1 or numbers.max() >= len(table)):
        raise ValueError(f"atomic numbers must be in 1..{len(table) - 1}, got {numbers.min()}..{numbers.max()}")
    return table[numbers]
```

aimnet/calculators/vibrations.py
```diff
@@ -0,0 +1,233 @@
...
+            returned, including the near-zero rigid-body ones.
+
+    Returns:
+        :class:`VibrationalAnalysis` with ``3N - n_tr_removed`` modes.
+    """
+    positions = np.asarray(positions, dtype=np.float64).reshape(-1, 3)
+    masses = np.asarray(masses, dtype=np.float64).reshape(-1)
+    n = len(masses)
+    if positions.shape[0] != n:
+        raise ValueError(f"positions has {positions.shape[0]} atoms but masses has {n}")
```
isayev, line 188: Minor input hygiene, same spirit as the Hessian path: `positions`/`masses` given as a `requires_grad` or CUDA tensor raise inside `np.asarray`, while `hessian` gets the `.detach().cpu().numpy()` treatment at `_as_square_hessian`. A shared `_to_numpy` helper for all three would make them consistent. And a finiteness/positivity check up front turns a `LinAlgError` deep in `eigh` into a clear message:

```python
if not np.all(np.isfinite(masses)) or np.any(masses <= 0):
    raise ValueError("masses must be finite and positive")
if not np.isfinite(h).all():
    raise ValueError("Hessian contains NaN or Inf")
```

aimnet/calculators/vibrations.py
```diff
@@ -0,0 +1,233 @@
...
+_AMU = 1.66053906660e-27  # unified atomic mass unit, kg
+_C = 299792458.0  # speed of light in vacuum, m/s
+
+# hbar*omega in eV for a mass-weighted Hessian eigenvalue omega^2 in eV A^-2 amu^-1.
+_EV_PER_SQRT_EIGENVALUE = _HBAR * 1e10 / math.sqrt(_E * _AMU)
+# Energy of a photon with wavenumber 1 cm^-1, in eV (``ase.units.invcm``).
+_INVCM = 100.0 * 2.0 * math.pi * _HBAR * _C / _E
+
+
+@dataclass
```
isayev, line 46: With ndarray fields, `a == b` raises (ambiguous truth value) and instances are unhashable but mutable. Freezing and dropping the generated `__eq__` avoids both:

```suggestion
@dataclass(frozen=True, eq=False)
```

aimnet/calculators/vibrations.py
```diff
@@ -0,0 +1,233 @@
...
+# CODATA 2018 recommended values (https://physics.nist.gov/cuu/Constants/).
+# e, h and c are exact since the 2019 SI revision.
+_E = 1.602176634e-19  # elementary charge, C
+_HBAR = 6.62607015e-34 / (2.0 * math.pi)  # reduced Planck constant, J s
+_AMU = 1.66053906660e-27  # unified atomic mass unit, kg
+_C = 299792458.0  # speed of light in vacuum, m/s
+
+# hbar*omega in eV for a mass-weighted Hessian eigenvalue omega^2 in eV A^-2 amu^-1.
+_EV_PER_SQRT_EIGENVALUE = _HBAR * 1e10 / math.sqrt(_E * _AMU)
+# Energy of a photon with wavenumber 1 cm^-1, in eV (``ase.units.invcm``).
```
isayev, line 42: Nit: `ase.units.invcm` is CODATA 2014 by default; this value equals `ase.units.create_units("2018")["invcm"]` exactly (the two differ by 8e-9 relative). The constant is right — the attribution isn't quite.

```suggestion
# Energy of a photon with wavenumber 1 cm^-1, in eV (``ase.units.create_units("2018")["invcm"]``).
```

tests/test_vibrations.py
```diff
@@ -0,0 +1,170 @@
...
+
+
+def test_translation_rotation_basis_is_orthonormal():
+    masses = masses_amu(WATER_NUMBERS)
+    basis = translation_rotation_basis(WATER, masses, is_linear=False)
+    assert basis.shape == (6, 9)
+    np.testing.assert_allclose(basis @ basis.T, np.eye(6), atol=1e-10)
+    co2_basis = translation_rotation_basis(CO2, masses_amu(CO2_NUMBERS), is_linear=True)
+    assert co2_basis.shape == (5, 9)
+    np.testing.assert_allclose(co2_basis @ co2_basis.T, np.eye(5), atol=1e-10)
```
isayev, line 91: This asserts `B Bᵀ = I`, never that `B` spans the rigid-body space — which is exactly what breaks (see the comment on the basis). And every projected test uses a mass-weighted Hessian equal to `k·I`, for which *any* subspace of the right dimension gives eigenvalue `k`, so the CO2 and water tests would pass with a completely wrong basis; the negative-curvature test builds its Hessian from the module's own basis, so it's circular. The one independent check is the diatomic, whose stretch lives in *x* while the contaminated columns are *y/z*.

A regression test that would have caught it: an *x*-aligned linear triatomic with ~1e-6 Å off-axis noise, asserting that each raw non-axial rigid vector is inside `span(B)`:

```python
def test_basis_spans_rotations_of_a_noisy_axis_aligned_linear_molecule():
    rng = np.random.default_rng(0)
    pos = np.array([[-1.16, 0, 0], [0, 0, 0], [1.16, 0, 0]]) + 1e-6 * rng.standard_normal((3, 3))
    masses = masses_amu(CO2_NUMBERS)
    basis = translation_rotation_basis(pos, masses, is_linear=True)
    sqm = np.sqrt(masses)
    r = pos - (pos * masses[:, None]).sum(0) / masses.sum()
    for axis in (np.array([0, 1.0, 0]), np.array([0, 0, 1.0])):   # the two non-axial rotations
        v = (np.cross(axis, r) * sqm[:, None]).ravel()
        assert np.linalg.norm(v - basis.T @ (basis @ v)) < 1e-10 * np.linalg.norm(v)
```

plus a frequency test with a *structured* (non-isotropic) bend block — e.g. build `H_mw` with distinct eigenvalues on an SVD-derived vibrational subspace — asserting the two bends come out degenerate for the molecule along x, y and z.

tests/test_vibrations.py
```diff
@@ -0,0 +1,170 @@
...
+    for candidate in (hessian.reshape(3, 3, 3, 3), torch.tensor(hessian.reshape(3, 3, 3, 3))):
+        result = analyze_hessian(candidate, WATER, masses)
+        np.testing.assert_allclose(result.frequencies_cm1, reference.frequencies_cm1)
+    with pytest.raises(ValueError, match="shape"):
+        analyze_hessian(np.eye(6), WATER, masses)
+    with pytest.raises(ValueError, match="atoms"):
+        analyze_hessian(hessian, WATER[:2], masses)
+
+
+def test_vibrational_analysis_wraps_calculator_eval():
```
isayev, line 155: This is the only test of the function most users will call, and it never touches the calculator. A `weights`-marked (and `ase`-marked) test on a real registry model — water with `AIMNet2Calculator("aimnet2")`, cross-checked against `ase.vibrations.VibrationsData` for the top three modes — runs in well under a second once weights are cached and would pin the real Hessian shape/dtype hand-off. I ran exactly that: 1694.2 / 3753.3 / 3873.2 cm^-1, identical to `VibrationsData.get_frequencies()`.

docs/tutorials/geometry_optimization.md
```diff
@@ -198,6 +198,21 @@ else:
...
+    ```python
+    from aimnet.calculators.vibrations import analyze_hessian, masses_amu
+
+    vib = analyze_hessian(result["hessian"], aspirin.get_positions(), masses_amu(numbers))
+    print(vib.frequencies_cm1)  # shape (3N - 6,), ascending; negative values are imaginary
+    print(vib.n_imaginary, vib.is_linear, vib.n_tr_removed)
+    modes = vib.modes  # unit-norm Cartesian normal modes, shape (n_modes, N, 3)
+    ```
+
+    `analyze_hessian` accepts the `(N, 3, N, 3)` tensor returned with `hessian=True` or the `(3N, 3N)` matrix from `AIMNet2ASE.get_hessian`, both in eV/Å^2 with positions in Å, and needs only numpy (`atoms.get_masses()` works in place of `masses_amu`). `vibrational_analysis(base_calc, data)` runs the Hessian call and the analysis in one step. The positive entries of `vib.energies_ev` can be passed straight to `IdealGasThermo(vib_energies=...)` in Step 5.
```
isayev, line 214: Two problems in this paragraph, one of them a silent thermochemistry error.

`IdealGasThermo` (ASE 3.27) sorts by `|ε|` and slices the last `3N-6` entries, so a `3N-7` list of "positive entries" is accepted without complaint and gives the wrong ZPE; a negative entry passes `_clean_vib_energies` (it only rejects complex) and makes `get_entropy` return nan. The safe rule is: proceed only if `n_imaginary == 0`, then pass *all* of `energies_ev`, with `geometry` taken from `is_linear`. Also `data` is not defined in this tutorial — the recipe above passes the dict inline.

```suggestion
    `analyze_hessian` accepts the `(N, 3, N, 3)` tensor returned with `hessian=True` or the `(3N, 3N)` matrix from `AIMNet2ASE.get_hessian`, both in eV/Å^2 with positions in Å, and needs only numpy (`atoms.get_masses()` works in place of `masses_amu`). `vibrational_analysis(base_calc, {"coord": coords, "numbers": numbers, "charge": 0})` runs the Hessian call and the analysis in one step. For Step 5, proceed only when `vib.n_imaginary == 0`, then pass all of `vib.energies_ev` to `IdealGasThermo(vib_energies=...)` with `geometry="linear" if vib.is_linear else "nonlinear"` — do not drop or take the absolute value of imaginary modes.
```

Step 5 itself (further down) still teaches the old recipe: it converts imaginary modes with `abs(freq)`, can drop genuine low modes with the `> 10` filter, and hard-codes `geometry="nonlinear"`. With this helper in the tutorial the two recipes disagree; I'd migrate Step 5 to the helper rather than annotate it. Your characterization of the existing Step 4 recipe is accurate — planar NH3 on the real model gives the helper `n_imaginary=1` at -1002.8 cm^-1 while the recipe prints "confirmed minimum".

isayev (changes requested):
Thanks for this — it's a genuinely useful helper, the physics core is right, and the ASE-free numpy implementation is the correct call (`ase.vibrations.VibrationsData` does *not* project rotations; on the same water Hessian it reports spurious 232i/205i/189i modes, so this is a proper superset, not a reimplementation).

I ran it against the real models rather than only reading it. **What checks out:** the CODATA conversion matches an independent derivation to 2e-16 (521.4709 cm^-1 per sqrt(eV Å^-2 amu^-1), hbar·omega = 0.0646542 eV); symmetrize → mass-weight → project → re-symmetrize → `eigh` is the right order; `modes` are properly Cartesian; the float32 calculator Hessian is fine for this (≤5e-4 cm^-1 run-to-run on the lowest modes, 0.02–0.16 cm^-1 against central differences of the calculator's own forces, projected rigid-body eigenvalues ≤7e-15 against a smallest kept 2.8e-2); no input mutation, no graph retention.

**What blocks:** the rigid-body basis is wrong for any axis-aligned linear molecule that isn't *exactly* linear, and the result depends on the lab orientation — details and a verified three-line fix on the QR block. The test suite is structurally unable to see it (isotropic fixtures, orthonormality asserted but not span), so the fix needs to come with the regression test sketched on the test file.

**Also needed before merge:**
- A `CHANGELOG.md` entry under `[Unreleased] > Added` — every user-visible change gets one here.
- The `IdealGasThermo` sentence in the docs tip (silently wrong ZPE with imaginary modes present); Step 5 further down still teaches the old recipe and now disagrees with the tip.
- A warning in `vibrational_analysis` for embedded-dispersion artifacts, whose Hessians currently lack the dispersion block (calculator-side issue on my list; the helper is where users will hit it). If you'd rather leave that to me, say so in the PR and I'll track it.

The remaining inline items (mass/Z validation, single-atom case, dataclass, tensor hygiene, the `create_units("2018")` attribution) are small and I've given suggestions where a line replacement is clean. Happy to re-review quickly once the basis change and its test are in.

isayev:
Applied the changes from my review directly on this branch so it can land — a week without a reply, and the fix was already verified end to end. On top of your commit:

- `translation_rotation_basis` builds the rigid-body basis with an SVD instead of unpivoted QR, with `is_linear_molecule` as the single rank authority. A near-linear molecule along *x* or *y* now gives the same frequencies as along *z* (real-model CO2: `[680.2 680.2 1373.7 2401.3]` in every orientation, where the QR version dropped a bend or produced a spurious mode). A single atom yields zero modes.
- Tests that can see it: span checks on a noisy axis-aligned linear molecule per axis, an orientation-independence test with a structured Hessian, a Wilson B-matrix/GF reference for a linear O-C-O and for a bent triatomic (derived from internal force constants, independent of the module), and a `weights`/`ase` cross-check of `vibrational_analysis` against `ase.vibrations.VibrationsData` on a real calculator.
- Input validation (mass table range, non-finite Hessian or masses, batch and `mol_idx` guards), tensor inputs accepted for every argument, a frozen result, and a warning when a model embeds a tabulated DFT-D3 module — that module's curvature is currently missing from the calculator's Hessian, which is a calculator-side issue I'll handle separately.
- `linear_tol` is exposed through `analyze_hessian` and `vibrational_analysis`, and its docstring now states what each misclassification costs; the default is unchanged.
- Docs: the tutorial's Step 5 uses the helper, gates on `n_imaginary == 0`, passes all `energies_ev` to `IdealGasThermo`, and takes `geometry` from the result; the module is in the API reference; CHANGELOG entry added.

All CI lanes are green. Thank you for the contribution — the projected analysis is a real improvement over the recipe it replaces, and it is now the documented path.

isayev (approved):
Review changes applied on the branch and verified end to end (SVD rigid-body basis with independent GF references, input validation, scoped dispersion warning, docs and CHANGELOG); CI green across the matrix. Approving to supersede my earlier request-changes.

PR #135, opened 2026-09-15
feat(aimnet2): compile inference and derivative training

zubatyuk:
## Summary

- Make AIMNet2 neighbor modes 0, 1, and 2 compatible with compiled masking, atom-count metadata, and molecular reductions.
- Add opt-in single-GPU compiled energy, force, and stress training with live model parameters, dynamic atom and neighbor counts, and a fixed input layout per trainer.
- Keep the calculator's original model as the parameter and state owner while routing ordinary CUDA inference through a fullgraph-compiled callable and higher derivatives through the eager model.
- Raise the minimum PyTorch version to 2.10 and align CI, GPU validation, and user documentation with the supported 2.10-2.14 range.
- Extend tests for compiled inference, training updates, loss masking, stress, Hessians, and Hessian-vector products.


aimnet/train/utils.py
```diff
@@ -216,19 +220,152 @@ def prepare_batch(batch: dict[str, Tensor], device="cuda", non_blocking=True) ->
...
+        # changing Torch's process-global compiler state.
+        with nbops._symbolic_trace_context():
+            traced = make_fx(
+                derivative_forward,
+                tracing_mode="symbolic",
+                decomposition_table=core_aten_decompositions(),
+                _allow_non_fake_inputs=True,
+                _error_on_data_dependent_ops=True,
+            )(*example_args)
+        self.forward = torch.compile(traced, dynamic=True, fullgraph=False)
```
isayev: **Blocking — the traced graph is built once and never re-validated, so any size-1 dimension is baked in as a literal.**

`make_fx(tracing_mode="symbolic")` records shape guards in its `ShapeEnv`, but it returns a bare `GraphModule` that checks none of them, and `_SymbolicTrainingForward` is constructed exactly once (line 361) with no retrace path. Combined with PyTorch's 0/1-size specialization, every dimension that happens to be `1` in the first batch is frozen as a constant in the graph.

Reproduced on an L40S with this PR's own tiny model (torch 2.13.0+cu130):

| trace batch | later batch | result |
|---|---|---|
| mode 0, B=2, N=4 | B=2, N=6 | OK (max abs dE = 7e-9) |
| mode 0, **B=1** | B=2 | `TorchRuntimeError` in `aten.view.default` |
| mode 0, **B=3** | **B=1** | `TorchRuntimeError` — the trace-time B is baked into a view |
| mode 1, **n_mol=1** | n_mol=2 | **CUDA device-side assert** (`ScatterGatherKernel.cu:203`, index out of bounds) — the CUDA context is poisoned and the process has to die |

The mode-1 case comes from `nbops.py:497-498`: `torch.zeros(charge.shape[0]).scatter_add_(...)` with `charge.shape[0] == 1` duck-specialized to the literal `1`, so `mol_idx == 1` then indexes out of bounds.

On reachability, to be fair: `SizeGroupedSampler._samples_list` uses `np.array_split` and `merge_groups(min_size=8*batch_size)`, so B=1 on the shipped dense path is a corner case (single-member size group, `batch_mode: atoms` with `N >= batch_size`, or a custom loader). What makes this blocking is the failure mode rather than the frequency — an opaque `aten.view` error in mode 0, and a poisoned CUDA context in mode 1 that can also take out unrelated work on the same device.

`__call__` (line 305) only compares the key *set*, though the class docstring promises that ranks, dtypes and neighbor mode are fixed too. Suggested fix: record a per-input signature of `(rank, dtype, tuple(d == 1 for d in shape))` at trace time and either cache one `GraphModule` per signature (retrace on miss) or raise a clear `ValueError`.

Note the parity test can't catch any of this — it feeds `[batch, reordered_batch]`, which is the same shapes twice with only the dict key order changed.

aimnet/train/utils.py
```diff
@@ -216,19 +220,152 @@ def prepare_batch(batch: dict[str, Tensor], device="cuda", non_blocking=True) ->
...
+    def __init__(self, model: nn.Module, example: dict[str, Tensor], target_keys: tuple[str, ...]):
+        self.module = model.module if isinstance(model, Forces) else model
+        self.input_keys = tuple(example)
+        self.state_names = tuple(name for name, _ in (*self.module.named_parameters(), *self.module.named_buffers()))
+        self.force_key = model.key_out if isinstance(model, Forces) else None
+        self.coord_key = model.x if isinstance(model, Forces) else "coord"
+        self.energy_key = model.y if isinstance(model, Forces) else "energy"
+        self.need_stress = "stress" in target_keys
+        if self.need_stress and "cell" not in example:
+            raise ValueError("Compiled training stress targets require a cell input.")
```
isayev: **Blocking — stress is gated on `cell` alone, so a mode-0 batch trains against a gas-phase cluster virial.**

`ops.calc_distances` uses `cell` *only* to turn `shifts` into Cartesian image displacements (`if f"shifts{suffix}" in data: coord_j = coord_j + shifts`). Mode 0 has no neighbor matrix and can never carry `shifts`, so with a dense batch the strained `cell` never enters the energy at all. `dE/dstrain` then collapses to the isolated-cluster virial and gets divided by an arbitrary `abs(det(cell))`.

The calculator never hits this because `mol_flatten` force-flattens to mode 1 whenever a cell is present and `make_nbmat` builds the shifts. The new training path has no equivalent guard, and mode 0 is exactly what the only shipped loader (`aimnet/data/sgdataset.py` — no `mol_idx`/`nbmat` anywhere in it) emits. So the default path for `trainer.compile` + stress regresses every label against a gas-phase virial; for a molecular crystal the image contribution is the dominant term, so this isn't a small bias.

Two reasons the tests can't see it:
1. The cell-strain path is dead in every fixture — modes 0/1 pass `cell` with no `shifts` at all, and mode 2 passes `shifts = torch.zeros(...)`, so `d(shifts @ cell)/d(strain)` is identically zero.
2. The `EagerStress` reference re-implements the formula under test (same einsum, same `cell @ strain`, same detached `det`), so a transposition, a sign flip or a wrong volume would pass too.

Suggested fix: require `shifts` (and reject mode 0) when `"stress" in target_keys`, here and in `default_evaluator`; and add one finite-difference check (`E(cell·(I+eps·e)) - E(cell)) / (eps·V)`) or a cross-check against `AIMNet2Calculator(...).eval(stress=True)` on a real periodic cell.

For what it's worth, the stress math itself is correct — I checked the sign, the row-vector convention on both `coord @ strain` and `cell @ strain`, the `abs(det(cell))` volume and the `.detach()` before the determinant against `derivatives.set_grad_tensors` / `get_derivatives`, and they all agree. It's only the missing periodicity precondition.

aimnet/nbops.py
```diff
@@ -467,21 +490,31 @@ def calc_masks(data: dict[str, Tensor]) -> dict[str, Tensor]:
...
-        # so mol_sum does not need a device-to-host sync on every call. Not
-        # cached under torch.compile: materializing bincount's data-dependent
-        # shape as a tensor inside the traced graph trips inductor codegen,
-        # and compiled mol_sum does not read the cache.
-        if not torch.compiler.is_compiling():
+        if _is_compiling() and "charge" in data:
+            # ``charge`` has one entry per molecule, so it supplies a fixed
+            # output size for the compiled reduction.  The final flat atom is
+            # padding and belongs to the final molecule.
+            mol_sizes = torch.zeros(data["charge"].shape[0], device=data["mol_idx"].device, dtype=torch.long)
```
isayev, line 499: The compiled branch and the eager branch disagree about **which molecule owns the padding atom**. Eager (`bincount` then `mol_sizes[-1] -= 1`) subtracts from the last *occupied* molecule; this branch subtracts from the last *declared* one (`charge.shape[0] - 1`).

```
charge=zeros(3), mol_idx=[0,0,0,1,1,1,1]   # 7th atom is padding
eager    mol_sizes -> tensor([3, 3])
compiled mol_sizes -> tensor([ 3,  4, -1])   # -> _natom**0.5 = [1.73, 2.00, nan]
```

Molecule 1 counts the padding atom as real and molecule 2 gets -1, so `energy_loss_fn`'s `_natom ** 0.5` yields NaN and every per-atom normalization downstream is wrong. The comment above asserts "`charge` has one entry per molecule", but nothing enforces it — `AIMNet2Base.prepare_input` only asserts `charge.ndim == 1`.

The shipped loader is mode-0 only so this needs a custom mode-1 loader, but note the repo's own `test_packed_dict_without_nbmat` uses the other packed convention (`mol_idx=[0,0,0,1,1,2]`, pad in its own bucket) — under that layout `scatter_add_` writes index `n_mol` into a size-`n_mol` tensor, which per this file's own mode-2 contract note is a CUDA device-side assert rather than a clean error.

Cheapest fix: assert `int(mol_idx.max()) + 1 == charge.shape[0]` once, eagerly, before tracing; or subtract the pad atom via `mol_idx[-1]` instead of a positional one-hot.

Unrelated nit on line 498: `torch.ones_like(mol_idx)` makes `scatter_add_` reject an `int32` `mol_idx` that eager `bincount` accepts (`self.dtype` must equal `src.dtype`). Unreachable through the model since `_prepare_dtype` casts to int64, but several test fixtures build these dicts by hand. `torch.ones_like(mol_idx, dtype=torch.long)` covers it.

aimnet/train/loss.py
```diff
@@ -63,12 +63,31 @@ def peratom_loss_fn(y_pred: dict[str, Tensor], y_true: dict[str, Tensor], key_pr
     x = y_true[key_true]
     y = y_pred[key_pred]
 
-    if y_pred["_natom"].numel() == 1:
-        loss = torch.nn.functional.mse_loss(x, y)
-    else:
+    numbers = y_pred.get("numbers")
+    nbmat = y_pred.get("nbmat")
+    if (
```
isayev: These new branches make `peratom_loss_fn` return a **materially different loss for the same physical batch** depending on which neighbor layout the loader produced. Measured with two systems (3 and 2 real atoms, uniform force error 0.1):

| layout | loss |
|---|---|
| mode 0, padded | 1.354e-3 |
| mode 1 (`mse(x[:-1], y[:-1])`) | 1.000e-2 |
| mode 2 (`mse` over `numbers != 0`) | 1.000e-2 |

That's a 7.4x spread. Since `MTLoss` normalizes component weights to sum to 1, a 7x shift in the forces component is a 7x shift in the effective force/energy weighting — i.e. the same config and the same data train a different model depending on the batch layout.

To be clear this is **new capability, not a regression**: `_natom` was only ever written in mode 0 on `main`, so modes 1 and 2 previously raised `KeyError` through this function. And the *pre-existing* mode-0 form is arguably the odd one out — `diff2 * (_natom/dim)` multiplies by the atom count rather than dividing, so even a fully-real batch returns `mse/3` for a `(B,N,3)` force tensor. The two new masked-MSE branches agree with mode-0 *unpadded* and are the defensible per-atom normalization, so I'd unify on masked MSE rather than make modes 1/2 reproduce the padded formula. Either way it needs a CHANGELOG entry, since it changes what existing configs optimize.

Separately, the branch selection here is shape duck-typing rather than a mode query. The root cause is at the seam: `Forces.forward` returns the full data dict (so eager consumers see `_nb_mode`), but `_SymbolicTrainingForward.__call__` returns only `dict(x)` plus `output_keys`, which drops `_nb_mode` and `mol_sizes`. Widening `output_keys` to carry `_nb_mode` would let this call `nbops.get_nb_mode` instead of guessing. As written, the guard on line 74 (`x.shape[:2] == numbers.shape`) can also match a `(B, 3)` per-molecule dipole against a `(B, N=3)` `numbers` — harmless for an unpadded batch, but it would drop dipole *components* as if they were padding atoms.

aimnet/train/utils.py
```diff
@@ -240,16 +377,44 @@ def _update(engine: Engine, batch: tuple[dict[str, Tensor], dict[str, Tensor]])
...
+            .unsqueeze(0)
+            .repeat(n_systems, 1, 1)
+            .requires_grad_(True)
+        )
+        if coord.ndim == 2:
+            x["coord"] = torch.einsum("ni,nij->nj", coord, strain[x["mol_idx"]])
+        else:
+            x["coord"] = torch.einsum("bni,bij->bnj", coord, strain)
+        x["cell"] = x["cell"] @ strain
+        y_pred = module(x)
```
isayev: The `stress=True` branch never re-establishes `torch.no_grad()`. The `stress=False` branch above keeps it (line 392), but this path calls `module(x)` directly and backwards through it on line 413, so every validation batch retains full activations.

`default_train.yaml` sets the validation sampler to `batch_size: 1024`, roughly 2x the training batch, so a compiled-stress run can OOM during validation on a model that trains fine. Either chunk the stress validation down to the training batch size, or document that `samplers.val.kwargs.batch_size` must be lowered when `trainer.compile` + stress is enabled.

Also, this block hardcodes `x["coord"]` and `y_pred["energy"]` while `_SymbolicTrainingForward` uses `self.coord_key = model.x` and `self.energy_key = model.y`. `Forces.__init__` accepts custom `x`/`y`, so a non-default wrapping would train on one pair of keys and validate on another. These two implementations of the same physics have already drifted — worth extracting one helper, or better a `Stress`/`Virial` module next to `Forces` in `modules/core.py` that both paths (and `EagerStress` in the tests) call.

aimnet/train/utils.py
```diff
@@ -216,19 +220,152 @@ def prepare_batch(batch: dict[str, Tensor], device="cuda", non_blocking=True) ->
...
+            state_tensors = tensors[len(self.input_keys) :]
+            data = dict(zip(self.input_keys, input_tensors, strict=True))
+            coord = data[self.coord_key]
+            need_derivatives = self.force_key is not None or self.need_stress
+            if need_derivatives:
+                coord = coord.detach().requires_grad_(True)
+                data[self.coord_key] = coord
+            strain = None
+            if self.need_stress:
+                n_systems = data["cell"].shape[0] if coord.ndim == 2 else coord.shape[0]
```
isayev: `n_systems = cell.shape[0]` assumes `cell` is always `(B, 3, 3)` in the flat branch, but a shared `(3, 3)` cell is a supported layout — `derivatives.set_grad_tensors` branches explicitly on `cell.ndim == 2` and builds a single `(3,3)` scaling for it.

With a shared cell here, `n_systems` becomes 3, so `strain` is `(3,3,3)` and `strain[mol_idx]` indexes molecules into a 3-row tensor: out of bounds for 4+ molecules, and silently the *wrong* per-molecule strain for 2-3 molecules (each row is identity so the value survives at eps=0, but the derivative is aliased across molecules, giving a stress summed over molecules sharing a row and divided by one molecule's volume). `cell @ strain` also broadcasts `(3,3) @ (3,3,3)` without complaint.

Mirroring the three-way `cell.ndim`/`coord.ndim` dispatch already in `derivatives.set_grad_tensors` — or simply rejecting a 2-D `cell` on this path — would close it. Same line exists at 400 in `default_evaluator`.

aimnet/train/utils.py
```diff
@@ -10,7 +10,11 @@
 from ignite.handlers import ModelCheckpoint, ProgressBar, TerminateOnNan, global_step_from_engine
 from omegaconf import OmegaConf
 from torch import Tensor, nn
+from torch._decomp import core_aten_decompositions
```
isayev: These two are private/experimental torch APIs imported at **module scope**, so they break all of `aimnet.train`, not just compiled training.

`torch._decomp` is private and `torch.fx.experimental` is explicitly experimental. With an unbounded `torch>=2.10` pin plus the scheduled `torch-latest` early-warning lane, an upstream rename turns `import aimnet.train.utils` into an `ImportError` for every user, including everyone running `compile: false`. Moving both inside `_SymbolicTrainingForward.__init__` confines the blast radius to the feature that actually needs them.

Related: I verified the `make_fx` signature only on the locked 2.13 —

```
(f, decomposition_table=None, tracing_mode='real', _allow_non_fake_inputs=False, *,
 pre_dispatch=False, ..., _error_on_data_dependent_ops=True, ...)
```

All four kwargs used here exist with the expected semantics there, but two are underscore-prefixed and `_allow_non_fake_inputs` is positional-or-keyword, so a signature reshuffle on 2.10/2.11/2.12/2.14 would be a silent `TypeError` at the first compiled step — and CI can't reach any of those (see the summary). Worth a comment recording the assumption. (`_error_on_data_dependent_ops=True` is already the default, so it documents rather than protects.)

aimnet/train/train.py
```diff
@@ -106,8 +106,14 @@ def run(local_rank, world_size, model_cfg, train_cfg, load, save):
...
 
+    compile_training = bool(train_cfg.trainer.get("compile", False))
+    if compile_training and world_size > 1:
+        raise RuntimeError("trainer.compile=True is not supported with DDP (world_size must be 1).")
+    if compile_training and not torch.cuda.is_available():
+        raise RuntimeError("trainer.compile=True requires a CUDA GPU.")
+
     # build model
-    _force_training = "forces" in train_cfg.data.y
+    _force_training = "forces" in train_cfg.data.y or (compile_training and "stress" in train_cfg.data.y)
```
isayev: This line makes `trainer.compile` — a performance switch — decide model construction, and it leaves stress training silently compile-only.

Nothing in `aimnet/modules/*` or `aimnet/models/*` produces a `"stress"` key; stress exists only in `_SymbolicTrainingForward.derivative_forward` and `default_evaluator(stress=True)`, and `build_engine` sets `stress=True` only when `compile` is on. So a config with `data.y: [energy, stress]` and `compile: false` builds fine, trains for exactly one batch, and then dies on `y_pred["stress"]` inside `mse_loss_fn` with no context pointing at the real cause.

An explicit check in `run()` — stress target without `trainer.compile` -> a `RuntimeError` naming both keys — would turn that into a config-time error. Separately, the `Forces` wrap is being used here purely as a carrier for `key_out`/`x`/`y` so `_SymbolicTrainingForward` can read them; passing those names explicitly would decouple the two concerns.

aimnet/train/utils.py
```diff
@@ -267,9 +432,16 @@ def __call__(self, engine):
 
 def build_engine(model, optimizer, scheduler, loss_fn, metrics, cfg, loader_val):
     device = next(model.parameters()).device
+    evaluator_name = cfg.trainer.evaluator
+    need_stress = bool(cfg.trainer.get("compile", False)) and "stress" in cfg.data.y
+    if need_stress and evaluator_name != "aimnet.train.utils.default_evaluator":
```
isayev: Gating the capability on string equality against a dotted import path is brittle: it rejects the identical callable reached through a re-export or an alias, and would accept a user module that happens to shadow that path. The resolved object is already available a few lines down — `get_module(evaluator_name) is default_evaluator` — or better, check the capability rather than the identity: `"stress" in inspect.signature(validate_fn).parameters`.

There's also an asymmetry with the trainer side. `train_kwargs["compile_training"] = True` (line 443) goes into `get_module(cfg.trainer.trainer)` with no equivalent check, so a third-party trainer without that kwarg gets a bare `TypeError` naming an argument the user never wrote (they wrote `trainer.compile: true`) — and one with `**kwargs` swallows it silently and trains eagerly while the user believes compilation is on. Since `trainer.trainer` and `trainer.evaluator` are documented extension points in `default_train.yaml`, both sides deserve the same signature check and the same clear message.

tests/test_ase.py
```diff
@@ -601,19 +601,29 @@ def test_hessian_nse_open_shell(self):
         # Sanity: doublet Hessian should be symmetric to fp32 noise.
         assert np.max(np.abs(H - H.T)) / np.max(np.abs(H)) < 1e-3
 
-    def test_hessian_compile_model_raises(self):
-        """compile_model=True must reject the Hessian path with RuntimeError."""
+    def test_hessian_compile_model_matches_eager(self):
```
isayev, line 605: This test needs `@pytest.mark.slow`. Measured on the locked env: **18.3 s in isolation**, 7.9 s warm inside the full suite — either way well over the 2 s threshold.

Every other test in `TestHessian` (lines 452, 467, 483, 500, 546, 576) carries the marker, and the test this replaces (`test_hessian_compile_model_raises`) returned before doing any compute. Unmarked, it is collected in both the local `make test` selection and CI's `tests-ase` lane (`-m "ase and not slow"`), where it's the only node this PR adds to the slow tail.

Same applies to `tests/test_calculator.py::test_compile_kwargs_parameter` (13.8 s isolated, unmarked, while its sibling `test_compile_model_parameter` directly above *is* marked) — pre-existing, but this PR touches that test's body so it's a cheap one to fold in.

isayev (commented):
## Review

I ran this on an 8x L40S box against the locked env (torch 2.13.0+cu130) as well as reading it. Summary of what the gates say:

- Targeted GPU suite (`test_compile_paths`, `test_train_utils`, `test_hvp`, `test_calculator`, `test_ase`): **270 passed, 0 failed** in 320 s.
- Default CPU lane: **670 passed, 13 skipped in 31.6 s** — inside our 2-minute budget.
- `quality` is red; everything else on CI is green.

The feature works. Compiled training runs end-to-end at 0.01-0.02 s/step with zero recompiles and zero graph breaks across a B x N sweep on the real `aimnet2.yaml` model, and I confirmed the removed hang guards are genuinely obsolete: with `compile_model=True` on CUDA, two warm `forces=True` calls (8.2 s then 0.046 s, force delta 1.2e-6 vs eager), then `hessian=True` (0.28 s, delta **exactly 0.0**), then `hessian_vector_product` (delta 7.6e-6), then forces again — no hang, exact agreement. `fullgraph=True` also compiled and ran cleanly on all four registry families (`aimnet2`, `nse`, `rxn`, `pd`). Nice work — the `_compiled_forward` design that keeps the original module as the parameter owner is the right call.

**The PyTorch 2.10 floor is fine, no objection** — please just give it its own `### Changed` entry rather than folding it into the "Added 2.14" bullet.

### Blocking

Two inline, both about the compiled training path:

1. **The traced graph is built once and never re-validated** (`train/utils.py`). Any size-1 dimension at trace time is baked in as a literal; a later batch that violates it gives an opaque `aten.view` error in mode 0 and a **CUDA device-side assert that poisons the context** in mode 1. Repro table inline.
2. **Stress is gated on `cell` alone with no `shifts` requirement** (`train/utils.py`). Mode 0 can never carry `shifts`, so the strained cell never enters the energy and stress labels get regressed against a gas-phase cluster virial. Mode 0 is what our only shipped loader emits, so this is the default path for `trainer.compile` + stress.

Neither is catchable by the current tests — the parity test feeds the same shapes twice with a reordered dict, and the cell-strain path is dead in every stress fixture (`shifts` is absent or all-zero), while `EagerStress` re-implements the very formula under test.

### Also worth fixing before merge

Inline: the mode-1 `mol_sizes` pad-atom convention mismatch (NaN `_natom`), the ~7.4x loss divergence across neighbor modes, the missing `no_grad` in stress validation, `n_systems` from `cell.shape[0]`, the module-level private torch imports, stress being silently compile-only, the string-matched evaluator gate, and a missing `slow` marker.

Three that aren't anchorable to a changed line:

- **`aimnet/ops.py:129`** is the one `torch.compiler.is_compiling()` call site not converted to `nbops._is_compiling()`. Under `_symbolic_trace_context()` it evaluates `False`, so the compiled-training graph takes the eager `index_select` branch and the documented inductor workaround is missing from exactly the graph inductor compiles — and the comment two lines down describes the alternative as a miscompile. It's inert today, but it goes live the moment the trace-once issue above is fixed with per-signature retracing (the workaround only fires when `F_u.shape[0] == 1` at trace time). I could not reproduce the original 2.9.1 miscompile on 2.13, so it may be worth re-validating rather than carrying forward. Same for the "inductor (2.9.1+cu128)" justifications at `nbops.py:726` and `ops.py:133` — 2.9 is out of the supported range now.
- **CHANGELOG** has no entry for compiled training, for Hessian/HVP now working under `compile_model=True` (removing those two documented `RuntimeError`s is a public contract change — anyone catching them has dead code), for the CUDA `fullgraph=True` default, for the `torch>=2.10` floor, or for `_natom` now being populated in modes 1/2.
- **`compile_model=True` is silently a no-op for legacy `.jpt` models** (`calculator.py:1031`): the `isinstance(self.model, torch.jit.ScriptModule)` branch is checked first and now wins, since `self.model` is no longer replaced by the Dynamo wrapper. The user pays the compile cost at `__init__` and gets nothing, with no warning.

### Two notes, no action needed

- `fullgraph=False` on the traced graph (line 296) reads as contradicting the PR description, but it's inert — a `make_fx` output has no Python control flow to break on, which is also why flipping it to `True` would cost nothing and make a future regression loud.
- The `torch.where` on a CPU `_input_padded` with CUDA operands in `loss.py` is fine: a 0-dim CPU condition is promoted as a scalar (a 1-dim one would raise), and `calc_masks` always writes 0-dim. Worth passing `device=` anyway, since our CPU-only CI never exercises the mixed-device combination.

### One process point

`tests-torch` runs on CPU runners with `TORCHDYNAMO_DISABLE=1`, compiled training raises without CUDA, and all six compiled-training tests are `@pytest.mark.gpu` — so **none of this surface is gated by CI**. `scripts/gpu_validate.sh` does run `pytest -m gpu` per torch version, so the manual matrix covers it, but only if someone runs `make gpu-validate`. Please note that dependency in the PR description, and consider adding a compiled-training leg to the script. Relatedly, no test currently exercises `_compiled_forward` with forces or stress — I instrumented it and the Hessian tests invoke it zero times, so `torch.compile` never actually runs in them, while four docs recommend `compile_model=True` specifically for force workflows.

Finally, `quality` is red on formatting only — `ruff format` wants `tests/test_ase.py`, `tests/test_hvp.py`, `tests/test_train_utils.py`, and `prettier` wants `.github/workflows/main.yml`, `README.md`, `docs/advanced/reaction_paths.md`. `make check` fixes all six.

Happy to talk through any of these, particularly the loss normalization question — that one is a genuine design call about what we want the objective to be, not just a bug.

bkalita23:
Putting autograd.grad(create_graph=True) inside make_fx, so Inductor only ever compiles a first-order backward, is the right approach, and passing live parameters through functional_call is clean. I ported _SymbolicTrainingForward onto an AIMNet2-based fork (torch 2.10.0+cu128, opt-einsum 3.4.0) and ran it on dense mode-0 batches from SizeGroupedDataset/SizeGroupedSampler. Our notes cover only the compiled training path; I have not looked at the calculator/inference changes yet.

1. With size-grouped data the batch shape changes almost every step, and that case is untested.
SizeGroupedSampler batches one size group at a time (np.array_split per group), so N changes from batch to batch and B ranges from 1 up to batch_size (small B for rare sizes). In a sample of 800 batches at 1,500 atoms per batch I had 302 distinct (B, N). The compiled force-training test runs [batch, reordered_batch] (one shape) and the stress test runs one batch. Could you add a test that trains across several (B, N), and neighbour counts for modes 1/2, and compares energy, forces and parameter gradients with eager at every step?

2. The symbolic trace can lock onto the first batch's shape. make_fx keeps no runtime shape checks, so a size decision taken on the first batch is baked in. I reproduced two ways (both failed loudly with a reshape error on a later batch):
- opt-einsum (in uv.lock via jax): for an einsum with 3+ operands, torch.einsum runs opt_einsum.contract_path on the symbolic sizes, which adds guards like Eq(s, B), Eq(s, N). Reproduced with the AEV contraction torch.einsum("...mag,...mdg,agh->...ahd", a, gv, agh). The current aev.py uses two-operand einsums, so today's model is not affected, but any model that uses one will be. Suggestion: compute the path on concrete sizes, check it is the same over the expected size range, and pass it explicitly (torch._VF.einsum(eq, ops, path=...)), or disable the path search inside _symbolic_trace_context.
- Duck sizing: equal sizes get one symbol. A first batch with B = 3 merged B with the xyz dimension, and a model that constrains that dimension (3×3 tensors) then pinned B. Small B is common with size-grouped sampling. Suggestion: torch.fx.experimental._config.use_duck_shape = False during the trace.
- Size 0/1 (not run): symbolic tracing specialises dimensions of size 1, so a first batch from a one-molecule group would presumably be fixed to B = 1.

Suggestion: after tracing, check gm.shape_env and refuse the graph if a batch or atom-count symbol was replaced by a constant, had its range narrowed, or appears in a guard that fails for other sizes. I implemented this check, and it caught a deliberately broken trace (B = 7, N = 17 pinned).

3. core_aten_decompositions() changed gradients. On 2-atom batches, in a model with an extra pairwise energy term with per-element embedding parameters, one parameter gradient was NaN with the decomposition table. Energy, forces and loss still matched eager, and gradients were finite both in eager and in the same symbolic trace without the table. Inductor applies its own decompositions, so the table may not be needed here. Worth adding gradient-parity tests on degenerate inputs (diatomics, masked or padded pairs).

4. No non-finite check before optimizer.step(). The step is clip_grad_value_(0.4) then step(), and TerminateOnNan only runs at EPOCH_COMPLETED. One non-finite gradient leaves NaN weights for the rest of the epoch. Suggestion: skip the update when the gradient norm is non-finite.

5. DDP. Compile is refused when world_size > 1 (train.py:110), while the default multi-GPU path is DDP, so large runs cannot use it. The refusal should stay: the compiled call goes through functional_call on the inner module and bypasses DDP.forward, so gradients would silently not sync. An explicit gradient all-reduce after loss.backward() would work with this forward.

6. Silent fallbacks. torch.compile(traced, dynamic=True, fullgraph=False): a graph break, or reaching Dynamo's recompile limit, falls back to eager without an error. For an already-traced graph, fullgraph=True plus logging or raising on recompiles would make that visible. (Not observed in our runs; this follows from how Dynamo behaves.)

7. Constraints to document. (a) Parameters and buffers are symbolic graph inputs, so model code that reads a buffer in Python (.item(), .tolist()) fails with GuardOnDataDependentSymNode. (b) __call__ checks only the input key names, not the ranks, dtypes or neighbour mode the docstring says must stay fixed.

8. Evidence at scale. Parity is tested on a 4-atom, nfeature=2 model for two SGD steps. Full-size parity over many steps with size-grouped batches, plus step time, compile time and peak memory against eager, would help. With the fixes for item 2 in place, the traced graph matched eager to about 1e-6 in forces and parameter gradients on 5 different batch shapes (CPU, uncompiled), apart from the NaN in item 3.

aimnet/train/utils.py
zubatyuk: `_CompiledTrainingRunner` now captures explicit symbols with duck shaping and singleton specialization disabled, and rejects incompatible contracts without a signature cache. CUDA coverage runs modes 0/1/2 through `B=1→2→3→1` and observes one forward and one backward compilation.

aimnet/train/utils.py
zubatyuk: Stress now rejects mode 0 and requires aligned shifts for every neighbor matrix in modes 1/2. Independent finite-difference tests cover both modes with nonzero periodic images.

aimnet/nbops.py
zubatyuk: Mode-1 counts now scatter `mol_idx[:-1]` into exactly `charge.shape[0]` long bins, with the dummy index normalized locally. Tests cover both dummy conventions and `int32` inputs.

aimnet/train/loss.py
zubatyuk: `peratom_loss_fn` now computes mean MSE over `numbers != 0` in every mode. The runner returns `numbers` explicitly instead of inferring the layout from shapes, and the mode-0 semantic change is documented as breaking.

aimnet/train/utils.py
zubatyuk: Eager and compiled force/stress evaluation now share the same derivative helper, removing the duplicated validation formula. Evaluator outputs are detached before metrics consume them.

aimnet/train/utils.py
zubatyuk: System count now comes from `charge.shape[0]`. `(3,3)`, `(1,3,3)`, and `(B,3,3)` cells are normalized before capture and replay; CUDA coverage verifies `B=2→3→2` with shared cells in modes 1/2.

aimnet/train/utils.py
zubatyuk: Private Torch imports now live inside `_capture`, and unsupported APIs raise a feature-local compatibility error. The decomposition table was removed, and Torch 2.10–2.14 were exercised.

aimnet/train/train.py
zubatyuk: `train.py` now selects the derivative runner whenever stress is a target, for both eager and compiled training. Stress no longer depends on a `Forces` carrier.

aimnet/train/utils.py
zubatyuk: `build_engine` now resolves the callables and checks evaluator stress support through its signature. `compile_training` is passed only to the resolved `default_trainer`; custom trainers remain responsible for their own compilation path.

tests/test_ase.py
zubatyuk: The Hessian and calculator compile tests are now marked `slow`. The three cited Python files are formatted, and Ruff lint and format checks pass.

zubatyuk:
@bkalita23, addressed each point in `a20692d`:

1. **“With size-grouped data the batch shape changes almost every step…”** The CUDA regression now runs `B=1→2→3→1` with changing atom counts, primary/LR neighbor widths, and `mult` in modes 0/1/2, checking eager/compiled E/F/loss/gradient/update parity. Real-data diagnostics also cover modes 0/1.
2. **“The symbolic trace can lock onto the first batch's shape…”** Capture now uses explicit symbols with `ShapeEnv(duck_shape=False, specialize_zero_one=False)` and `do_not_specialize_zero_one=True`; runtime checks reject incompatible contracts. Opt-einsum is unchanged because current AIMNet2 contractions have two operands.
3. **“core_aten_decompositions() changed gradients…”** `core_aten_decompositions()` was removed. The diatomic regression requires finite parameter gradients without that decomposition table.
4. **“No non-finite check before optimizer.step()…”** Loss is checked before backward and gradients before clipping or stepping. DDP all-reduces the failure flag so all ranks fail together.
5. **“DDP. Compile is refused when world_size > 1…”** `_CompiledTrainingRunner` is created before `idist.auto_model`, so DDP wraps the runner and every forward passes through DDP. The two-rank NCCL test covers changing local shapes, unused parameters, synchronized failures, matching updates, and no hang.
6. **“Silent fallbacks…”** Training now passes symbolic FX and fake inputs directly to AOTAutograd/Inductor, and the regression observes one forward and one backward compilation. Inference requires fullgraph compilation and rejects `fullgraph=False`.
7. **“Constraints to document…”** Runtime validation fixes ordered keys, ranks, dtypes, device type, neighbor mode, requested properties, and non-dynamic shapes. Private Torch imports are local to `_capture`, and unsupported APIs raise a feature-local compatibility error.
8. **“Evidence at scale…”** Real-data diagnostics cover fixed 5,120- and 10,240-atom batches in modes 0/1, with E/F/loss/gradient parity within established tolerances. Peak memory and strict AdamW trajectory equality were not measured.

zubatyuk:
@isayev, the non-inline review points are addressed in `a20692d`:

- **“The PyTorch 2.10 floor is fine…”** The minimum now has its own entry under `### Changed`.
- **“aimnet/ops.py:129 is the one torch.compiler.is_compiling() call site…”** Compiler checks now consistently use `nbops._is_compiling()`. The obsolete Torch 2.9 branches were removed, and the localized Torch 2.12 singleton workaround was revalidated without changing returned values.
- **“CHANGELOG has no entry…”** Separate entries now cover the Torch minimum, fullgraph inference, compiled E/F/stress training, DDP, eager stress, HVP/Hessian fallback, `_natom`/loss semantics, and `.jpt` behavior.
- **“compile_model=True is silently a no-op for legacy .jpt models…”** Requesting compilation for a legacy `.jpt` model now raises immediately.
- **“fullgraph=False on the traced graph…”** Training now passes symbolic FX and fake arguments directly to AOTAutograd/Inductor. Inference enforces fullgraph compilation and rejects `fullgraph=False`.
- **“The torch.where on a CPU _input_padded with CUDA operands…”** That branch was removed; `peratom_loss_fn` now uses `numbers != 0`.
- **“None of this surface is gated by CI…”** `scripts/gpu_validate.sh` now installs `[ase,train]`, and its GPU selection includes compiled-training tests across the Torch matrix. This remains the explicit GPU validation path rather than the CPU lane.
- **“No test currently exercises _compiled_forward with forces or stress…”** Calculator E/F compile tests now exercise `_compiled_forward`; training stress is covered separately through `_CompiledTrainingRunner`.
- **“quality is red on formatting only…”** The three cited Python files were formatted, and Ruff lint and format checks pass.

